PR #1578, opened 2017-10-18
Initial CLI and file format for fleetctl

marpaia:
This PR implements a program called `fleetctl` which scaffolds a high level CLI interface which can be used to manage a Fleet server. Configuration is articulated using an intent-based API that resembles the API that is used to configure [Kubernetes clusters](https://www.mirantis.com/blog/introduction-to-yaml-creating-a-kubernetes-deployment/). The idea here is to use the Kubernetes file format as a pattern to reduce the need for operators to become too intimately familiar with dramatically different file formats.

## Help Text

```
$ make fleetctl
$ ./build/fleetctl --help
fleetctl controls an instance of the Kolide Fleet osquery fleet manager.

Find more information at https://kolide.com/fleet

  Usage:
    fleetctl [command] [flags]


  Commands:
    fleetctl query    - run a query across your fleet

    fleetctl apply    - apply a set of osquery configurations
    fleetctl edit    - edit your complete configuration in an ephemeral editor
    fleetctl config   - modify how and which Fleet server to connect to

    fleetctl help     - get help on how to define an intent type
    fleetctl version  - print full version information

```

## File Format

Consider the following file, which contains all information about your Fleet deployment:

```yaml
apiVersion: v1/alpha
kind: config
spec:
  config:
    - distributed_interval: 3
    - distributed_tls_max_attempts: 3
    - logger_plugin: tls
    - logger_tls_endpoint: /api/v1/osquery/log
    - logger_tls_period: 10
---
apiVersion: v1/alpha
kind: decorators
spec:
  decorators:
    - name: hostname
      query: select hostname from system_info;
      type: interval
      interval: 10
    - name: uuid
      query: select uuid from osquery_info;
      type: load
    - name: instance_id
      query: select instance_id from osquery_info;
      type: load
---
apiVersion: v1/alpha
kind: fim
spec:
  fim:
    interval: 500
    groups:
      - name: etc
        paths:
          - /etc/%%
      - name: users
        paths:
          - /Users/%/Library/%%
          - /Users/%/Documents/%%

---
apiVersion: v1/alpha
kind: labels
spec:
  labels:
    - name: all_hosts
      query: select 1;
    - name: macs
      query: select 1 from os_version where platform = "darwin";
    - name: ubuntu
      query: select 1 from os_version where platform = "ubuntu";
    - name: centos
      query: select 1 from os_version where platform = "centos";
    - name: windows
      query: select 1 from os_version where platform = "windows";
    - name: pending_updates
      query: SELECT value from plist where path = "/Library/Preferences/ManagedInstalls.plist" and key = "PendingUpdateCount" and value > "0";
      platforms:
        - darwin
    - name: slack_not_running
      query: >
        SELECT * from system_info
        WHERE NOT EXISTS (
          SELECT *
          FROM processes
          WHERE name LIKE "%Slack%"
        );
---
apiVersion: v1/alpha
kind: pack
metadata:
  name: osquery_monitoring
spec:
  queries:
    - name: osquery_schedule
      interval: 7200
      removed: false
      description: Report performance publisher health and track event counters.
    - name: osquery_events
      interval: 86400
      removed: false
      description: Report event publisher health and track event counters.
    - name: oquery_info
      interval: 600
      removed: false
      description: A heartbeat counter that reports general performance (CPU, memory) and version.
---
apiVersion: v1/alpha
kind: query
metadata:
  - name: processes
spec:
  query: select * from processes;
---
apiVersion: v1/alpha
kind: queries
spec:
  queries:
    - name: launcher_version
      query: select version from kolide_launcher_info;
    - name: osquery_version
      query: select version from osquery_info;
    - name: osquery_schedule
      query: select name, interval, executions, output_size, wall_time, (user_time/executions) as avg_user_time, (system_time/executions) as avg_system_time, average_memory, last_executed from osquery_schedule;
      description: Report performance publisher health and track event counters.
    - name: osquery_events
      query: select name, publisher, type, subscriptions, events, active from osquery_events;
      description: Report event publisher health and track event counters.
    - name: osquery_info
      query: select i.*, p.resident_size, p.user_time, p.system_time, time.minutes as counter from osquery_info i, processes p, time where p.pid = i.pid;
      description: A heartbeat counter that reports general performance (CPU, memory) and version.
```

## Possible Directory Structure

As this file gets larger and more complex, you could also separate each server intent into a file in the following structure:

```
|-- config.yml
|-- decorators.yml
|-- fim.yml
|-- labels.yml
|-- packs
|   `-- osquery-monitoring.yml
`-- queries
    |-- processes.yml
    `-- queries.yml
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,8 +97,16 @@ endif
 .pre-fleet:
 	$(eval APP_NAME = fleet)
 
-build: .prefix .pre-build .pre-fleet
+.pre-fleetctl:
+	$(eval APP_NAME = fleetctl)
+
+build: fleet
+
+fleet: .prefix .pre-build .pre-fleet
 	go build -i -o build/${OUTPUT} -ldflags ${KIT_VERSION} ./cmd/fleet
+
+fleetctl: .prefix .pre-build .pre-fleetctl
+	go build -i -o build/fleetctl -ldflags ${KIT_VERSION} ./cmd/fleetctl
 
 lint-js:
 	eslint frontend --ext .js,.jsx
--- a/cmd/fleetctl/fleetctl.go
+++ b/cmd/fleetctl/fleetctl.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/kit/logutil"
+	"github.com/kolide/kit/version"
+)
+
+func runVersion(args []string) error {
+	version.PrintFull()
+	return nil
+}
+
+func runNoop(args []string) error {
+	fmt.Printf("%+v\n", args)
+	return nil
+}
+
+type runFunc func([]string) error
+type subcommandMap map[string]runFunc
+type commandMap map[string]subcommandMap
+
+func usage() {
+	fmt.Fprintf(os.Stderr, "fleetctl controls an instance of the Kolide Fleet osquery fleet manager.\n")
+	fmt.Fprintf(os.Stderr, "\n")
+	fmt.Fprintf(os.Stderr, "Find more information at https://kolide.com/fleet\n")
+	fmt.Fprintf(os.Stderr, "\n")
+	fmt.Fprintf(os.Stderr, "  Usage:\n")
+	fmt.Fprintf(os.Stderr, "    fleetctl [command] [flags]\n")
+	fmt.Fprintf(os.Stderr, "\n")
+	fmt.Fprintf(os.Stderr, "\n")
+	fmt.Fprintf(os.Stderr, "  Commands:\n")
+	fmt.Fprintf(os.Stderr, "    fleetctl query    - run a query across your fleet\n")
+	fmt.Fprintf(os.Stderr, "\n")
+	fmt.Fprintf(os.Stderr, "    fleetctl apply    - apply a set of osquery configurations\n")
+	fmt.Fprintf(os.Stderr, "    fleetctl edit    - edit your complete configuration in an ephemeral editor\n")
+	fmt.Fprintf(os.Stderr, "\n")
+	fmt.Fprintf(os.Stderr, "    fleetctl help     - get help on how to define an intent type\n")
+	fmt.Fprintf(os.Stderr, "    fleetctl version  - print full version information\n")
+	fmt.Fprintf(os.Stderr, "\n")
+}
+
+func main() {
+	logger := level.NewFilter(log.NewJSONLogger(os.Stderr), level.AllowDebug())
+	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
+	logger = log.With(logger, "caller", log.DefaultCaller)
+
+	if len(os.Args) < 2 {
+		usage()
+		os.Exit(0)
+	}
+
+	var run func([]string) error
+	switch strings.ToLower(os.Args[1]) {
+	case "version":
+		run = runVersion
+	case "query":
+		run = runNoop
+	case "edit":
+		run = runNoop
+	case "new":
+		run = runNoop
+	case "apply":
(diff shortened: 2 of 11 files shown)


marpaia:
We spoke about using `kubectl` as a reference, but I also want to pay attention to `knife` which is similar in that it's a cli that allows a user to manage a directory of source controlled files and modify a server deployment (chef): https://docs.chef.io/knife.html

marpaia:
So my first attempt was, in my opinion, too much of an actual API client via the command-line and not as much of a tool that allows you to manage config files in source control. I reworked to usage to reflect more of a usage of `fleetctl new project` to set up your repo. following that, all edits should happen to files and then you can `fleetctl apply packs`, etc. (and check in to git). recreating your fleet deployment should be possible from the git repo alone. consider the following help:

```
fleetctl - A CLI for Fleet

  Usage:
    fleetctl [command] [flags]

  Commands:
    fleetctl version

    fleetctl run query --help

    fleetctl new project --help

    fleetctl help labels --help
    fleetctl help decorators --help
    fleetctl help fim --help
    fleetctl help options --help
    fleetctl help packs --help
    fleetctl help queries --help

    fleetctl apply decorators --help
    fleetctl apply fim --help
    fleetctl apply options --help
    fleetctl apply packs --help
    fleetctl apply queries --help
    fleetctl apply labels --help


```

cmd/fleetctl/main.go
```diff
@@ -0,0 +1,85 @@
...
+			version.PrintFull()
+			os.Exit(0)
+		}
+	}
+
+	logger := level.NewFilter(log.NewJSONLogger(os.Stderr), level.AllowDebug())
+	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
+	logger = log.With(logger, "caller", log.DefaultCaller)
+
+	commands := commandMap{
```
groob: A few notes notes about the command list: 
a) It's missing `GET` features. We need to be able to export a pack, which was a big motivation for the original request. 

b) I think the list is very ambitious. We should trim it down to a more basic list of commands we plan to implement and go from there. 

c) There needs to be something to configure the local keys. `fleetctl login` or `fleetctl config set` 

d) I think the operations should be plural, similar to the REST endpoints. That way we can leverage the list methods with filters to get one or more result. For example:
```fleetctl get queries -packs=packs,scheduled,in```

marpaia: A) I intentionally did not include `get` commands. My intention was that if you wanted to "get" a pack or query text, you should just cat the file (which is currently TBD). 

B) Do you see anything that has the potential to be removed / is less MVP than the others?

C) Agreed.

D) Well, there is no `fleetctl get`, so let's ignore that. Everything else is already plural.

groob: I don't understand what you mean by intentionally excluding GET. Can you explain how you query current queries/packs/users etc?

marpaia: All of these data model objects are in source controlled files. The fleetcl command merely allows you to sync the changes you made to these files to the server. If you want to "get" a pack on the command line, cat the file.

groob: I'm intrigued. This is a big departure from the current model, but I want to hear more of this in an RFC. 

Are you brainstorming a file format? 

Also even though kubernetes works in a similar way, you can still query the current state and dump it back. One of the reasons is that a user can modify a local file and apply it, but not check it in source control. There's no way to diff them if you can't query what's available server side.

marpaia: I could imagine `fleetctl sync` which would download current state and allow you to `git diff` to observe differences.

I have some ideas on file format, I need to write them up.

marpaia: Maybe `fleetctl pull` 🤔

marpaia:
I'm going to merge this and submit a follow-up PR with an addition to the docs/ directory outlining the ideas here.

PR #1583, opened 2017-10-23
add health package to create reusable healthz checks.

groob:
Moved the healthz handler and exposed the CheckHealth method so the same healthchecks could be
used by TLS, gRPC and any other APIs.

---
I refactored this into a new package for #1582 CheckHealth implementation.


groob:
Let's keep it in fleet for now as I'm not totally sold on the interface. If we end up needing this in another server we can move it IMO.

PR #1590, opened 2017-10-27
Set decorator names for default decorators

### LINKED ISSUE #1518: Decorator page should have default names for built-in decorators
Right now, the decorator page doesn't have a name for the built-in decorators and I cannot click into them to see what the complete queries are. Ideally, we should set a name like "UUID" and "Hostname" for them by default.

screen shot 2017-06-03 at 11 39 10 am

marpaia:
close #1518


marpaia:
@groob woops, added!

PR #1598, opened 2017-11-12
Use a working MySQL version

viq:
Use repo that installs `5.7` version of MySQL, as opposed to `5.6` that fails when running `prepare db`


marpaia:
Thank you for finding this bug and contributing this!

PR #1606, opened 2017-11-14
Fix issue rendering hosts with unexpected CPU brand string

### LINKED ISSUE #1604: All Host page (and others) not rendering hosts
Initially, when first using the app, the UI will load hosts properly when they're visiting the All Hosts pages or any label (online, darwin, etc. At some instant however, the app will stop showing hosts in the UI. From this point forward, no hosts are able to be shown in the UI as cards or in the list. The sidebar, however, stays correct. Consider the following screenshot. The sidebar indicates that 11 hosts are online, but they are not rendered on the page:

image

@groob has debugged this a bit and determined that the response from the server contains the correct host data and has speculated that a new version of a JS dependency may be causing the rendering bug.

No noticeable errors occur in the console when this occurs and there is not currently a reliable set of reproduction steps.

zwass:
- Hosts with unexpected CPU brand string render properly
- Other unexpected host CPU information renders properly
- Add tests for host CPU parsing

Fixes #1604

### DIFF
--- a/frontend/redux/nodes/entities/hosts/helpers.js
+++ b/frontend/redux/nodes/entities/hosts/helpers.js
@@ -14,6 +14,7 @@ export const parseEntityFunc = (host) => {
     } catch (e) {
       // Some CPU brand strings do not fit this format and we can't parse the
       // clock speed. Leave it set to 'Unknown'.
+      console.log(`Unable to parse clock speed from cpu_brand: ${host.cpu_brand}`);
     }
     if (host.cpu_physical_cores || clockSpeedOutput) {
       hostCpuOutput = `${host.cpu_physical_cores || 'Unknown'} x ${clockSpeedOutput || 'Unknown'} GHz`;
--- a/frontend/redux/nodes/entities/hosts/helpers.js
+++ b/frontend/redux/nodes/entities/hosts/helpers.js
@@ -4,14 +4,20 @@ export const parseEntityFunc = (host) => {
   const { network_interfaces: networkInterfaces } = host;
   const networkInterface = networkInterfaces && find(networkInterfaces, { id: host.primary_ip_id });
 
-  let clockSpeed = null;
-  let clockSpeedFlt = null;
   let hostCpuOutput = null;
-
-  if (host && host.cpu_brand) {
-    clockSpeed = host.cpu_brand.split('@ ')[1] || host.cpu_brand.split('@')[1];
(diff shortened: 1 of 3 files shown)


frontend/redux/nodes/entities/hosts/helpers.js
```diff
@@ -4,14 +4,20 @@ export const parseEntityFunc = (host) => {
...
 
-  let clockSpeed = null;
-  let clockSpeedFlt = null;
   let hostCpuOutput = null;
-
-  if (host && host.cpu_brand) {
-    clockSpeed = host.cpu_brand.split('@ ')[1] || host.cpu_brand.split('@')[1];
-    clockSpeedFlt = parseFloat(clockSpeed.split('GHz')[0].trim());
-    hostCpuOutput = `${host.cpu_physical_cores} x ${Math.floor(clockSpeedFlt * 10) / 10} GHz`;
+  if (host) {
```
groob, line 8: Would it be possible to just do away with the parsed variant, and to print the string as is?

If not, the current implementation LGTM

frontend/redux/nodes/entities/hosts/helpers.js
```diff
@@ -4,14 +4,20 @@ export const parseEntityFunc = (host) => {
...
+    let clockSpeedOutput = null;
+    try {
+      const clockSpeed = host.cpu_brand.split('@ ')[1] || host.cpu_brand.split('@')[1];
+      const clockSpeedFlt = parseFloat(clockSpeed.split('GHz')[0].trim());
+      clockSpeedOutput = Math.floor(clockSpeedFlt * 10) / 10;
+    } catch (e) {
+      // Some CPU brand strings do not fit this format and we can't parse the
```
groob, line 15: Should there be a console.log(e) here?

frontend/redux/nodes/entities/hosts/helpers.js
zwass: Not with the current design
screen shot 2017-11-13 at 7 35 44 pm

frontend/redux/nodes/entities/hosts/helpers.js
zwass: I had debated adding a console log, but thought it may end up fairly noisy. We probably should add it and 99% of users won't notice the noise.

zwass: This is done.

groob:
@zwass want to add this to CHANGELOG too?

Issue #1615, opened 2017-11-15
osquery logger fails to flush logs to file


groob:
If a user starts fleet with `--osquery_enable_log_rotation`, after a certain time, the log is not being written to disk. 
Setting the flag to false, removes this problem. Because the `lumberjack` logger is used with `osquery_enable_log_rotation` it leads me to think that the bug is related to the lumberjack logger. 

https://github.com/natefinch/lumberjack/issues/55 sounds suspicious. 

Edit: looks like the logwriter can fail to write to disk even if it's not lumberjack. I will see if I can repro this under load.

groob:
In debugging this further I see that the logwriter can't deal with a deleted `/tmp/osquery_result` file.

groob:
```
ts=2017-11-15T14:23:33.745897Z component=service method=AuthenticateHost err=null took=3.462124ms
ts=2017-11-15T14:23:33.747546Z component=service method=SubmitResultLogs err=null took=502.252µs
ts=2017-11-15T14:23:34.728435Z component=service method=AuthenticateHost err=null took=3.448368ms
ts=2017-11-15T14:23:34.730132Z component=service method=SubmitResultLogs err=null took=540.144µs
ts=2017-11-15T14:23:35.699733Z component=service method=AuthenticateHost err=null took=3.989732ms
ts=2017-11-15T14:23:35.701558Z component=service method=SubmitResultLogs err=null took=512.859µs

cat /tmp/osquery_result
cat: /tmp/osquery_result: No such file or directory
```

ilya1st:
you can try my writer(https://github.com/ilya1st/rotatewriter). It does not lock app on rotation. but there are no compression for rotated files. my first goal was handle SIGHUP correcty in one of my private projects.

PR #1617, opened 2017-11-15
logwriter: stat file before writing

groob:
If the file was rotated/deleted the data is never written to disk. Calling os.Stat ensures that the file
will be re-created.

For #1615 

This PR only fixes the issue if `--osquery_enable_log_rotation` is `false`, which happens to be the default value.

### DIFF
--- a/server/logwriter/logwriter.go
+++ b/server/logwriter/logwriter.go
@@ -3,10 +3,11 @@ package logwriter
 
 import (
 	"bufio"
-	"errors"
 	"io"
 	"os"
 	"sync"
+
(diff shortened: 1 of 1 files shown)


zwass:
Does this problem manifest only if the user moves/deletes the open log file?

groob:
Yes, but the server does not see that the file handle is no longer usable. Calling `logrotate` on `/tmp/osquery_result` would cause this issue to manifest. Same with `rm /tmp/osquery_result`. 

This change recreates the file before writing if it doesn't exist.

PR #1623, opened 2017-11-15
use a forwarder which writes result logs to oklog.

groob:
minimal experiment

```
oklog ingeststore -store.segment-replication-factor 1
./build/fleet serve --auth_jwt_key=foo

<schedule some pack queries>

# distributed tail -f
oklog stream

# query 
oklog query -from 10m -q 'some|regex'

### CODE REVIEW

**Comment by marpaia** [comment]
Will we bundle oklog with fleet here? How will the UX of collecting logs change for those who just want to log to the filesystem (if at all)?

**Comment by groob** [comment]
Long term maybe? right now the interface between the two is minimal. oklog is standalone and fleet writes to oklog using this new writer.

The UX for those writing to the filesystem should not change, but I have to expose the config via flags. 

Imagine something like this:

```
--osquery_result_loggers=filesystem,oklog
--oklog-peers=localhost,oklog.acme.co,oklog2.acme.co
```

fleet can use https://golang.org/pkg/io/#MultiWriter to write to n number of outputs.

**Comment by marpaia** [comment]
That would be awesome, I love the idea of multiwriter and independently configurable output sinks.

**Comment by groob** [comment]
```
osquery:
  node_key_size: 24
  status_log_file: /tmp/osquery_status
  result_log_file: /tmp/osquery_result
  enable_log_rotation: false
  label_update_interval: 1h0m0s
  status_log_writers:
  - filesystem
  result_log_writers:
  - filesystem
  oklog_ingesters:
  - localhost
```

marpaia:
So in the above example, you're configuring oklog but you're not using it, right?

groob:
Right, the `localhost` is just a default value. I'm writing the initialization right now, and each writer is only configured if it's available in the `status_log_writers` or `result_log_writers` slice respectively.

groob:
I gotta do a bunch of cleanup and fix tests, but this works and is configurable. I gotta take if for a spin with heroku.

groob:
I'm going to archive this PR for now. There's a few things that are still not quite there for oklog to be a viable output for fleet. 

- log lines are limited to 64k
- no good support for json format in query. 
- no TLS for the ingester
- not oklog's fault, but it's hard to enable discovery on a system like heroku
- oklog can already be used without this change by running `tail -f osquery_result | oklog forward localhost`

This integration might be cool in the future, but it doesn't yet make sense to merge in to master.

PR #1625, opened 2017-11-16
Make OsqueryOptions hierarchical by platform

zwass:
- Allow overriding base config on a per-platform basis.
- Merge FIM configs into the OsqueryOptions object.


groob:
Perhaps something that should be discussed separately, but configs need to be applied per host. Maybe with labels?

marpaia:
What about something like:

```
---
apiVersion: k8s.kolide.com/v1alpha1
kind: OsqueryOptions
spec:
  config:
    distributed_interval: 3
    distributed_tls_max_attempts: 3
    logger_plugin: tls
    logger_tls_endpoint: /api/v1/osquery/log
    logger_tls_period: 10
  # Note configs in overrides take precedence over configs defined above
  overrides:
    platforms:
      darwin:
        - disable_tables: chrome_extensions
        - docker_socket: /var/run/docker.sock
        - logger_tls_period: 60
      linux:
        - schedule_timeout: 60
        - docker_socket: /etc/run/docker.sock
```

This maintains the ability for us to `override` by additional things (`host`, `label`, etc) as the need arises.

marpaia:
Also, we have to account for FIM in here as well so if we just wanted fim on macOS, it would be more like:

```
---
apiVersion: k8s.kolide.com/v1alpha1
kind: OsqueryOptions
spec:
  config:
    distributed_interval: 3
    distributed_tls_max_attempts: 3
    logger_plugin: tls
    logger_tls_endpoint: /api/v1/osquery/log
    logger_tls_period: 10
  # Note configs in overrides take precedence over configs defined above
  overrides:
    platforms:
      darwin:
        disable_tables: chrome_extensions
        docker_socket: /var/run/docker.sock
        logger_tls_period: 60
        fim:
          interval: 500
          groups:
            - name: etc
              paths:
                - /etc/%%
            - name: users
              paths:
                - /Users/%/Library/%%
                - /Users/%/Documents/%%
      linux:
        schedule_timeout: 60
        docker_socket: /etc/run/docker.sock
```

PR #1626, opened 2017-11-16
Documenting deploying Fleet on Kubernetes

marpaia:
Included in this PR are new docs on installing Fleet on Kubernetes as well as a deployment file, a load balancer service file, and a database migration job file.

### DIFF
--- a/docs/infrastructure/fleet-on-kubernetes.md
+++ b/docs/infrastructure/fleet-on-kubernetes.md
@@ -1,0 +1,147 @@
+# Deploying Fleet on Kubernetes
+
+In this guide, we're going to install Fleet and all of it's application dependencies on a Kubernetes cluster. Kubernetes is a container orchestration tool that was open sourced by Google in 2014.
+
+## Installing Infrastructure Dependencies
+
+For the sake of this tutorial, we will use Helm, the Kubernetes Package Manager, to install MySQL and Redis. If you would like to use Helm and you've never used it before, you must run the following to initialize your cluster:
+
+```
+helm init
+```
+
+### MySQL
+
+The MySQL that we will use for this tutorial is not replicated and it is not Highly Available. If you're deploying Fleet on a Kubernetes managed by a cloud provider (GCP, Azure, AWS, etc), I suggest using their MySQL product if possible as running HA MySQL in Kubernetes can be difficult. To make this tutorial cloud provider agnostic however, we will use a non-replicated instance of MySQL.
+
+To install MySQL from Helm, run the following command. Note that there are some options that are specified. These options basically just enumerate that:
+
+- There should be a `kolide` database created
+- The default user's username should be `kolide`
+
+```
+helm install \
+  --name fleet-database \
+  --set mysqlUser=kolide,mysqlDatabase=kolide \
+  stable/mysql
+```
+
+This helm package will create a Kubernetes `Service` which exposes the MySQL server to the rest of the cluster on the following DNS address:
+
+```
+fleet-database-mysql:3306
+```
+
+We will use this address when we configure the Kubernetes deployment and database migration job, but if you're not using a Helm-installed MySQL in your deployment, you'll have to change this in your Kubernetes config files.
+
+#### Database Migrations
+
+The last step is to run the Fleet database migrations on your new MySQL server. To do this, run the following:
+
+```
+kubectl create -f ./examples/kubernetes/fleet-migrations.yml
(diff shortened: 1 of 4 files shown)


docs/infrastructure/fleet-on-kubernetes.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+We will use this address when we configure the Kubernetes deployment and database migration job, but if you're not using a Helm-installed MySQL in your deployment, you'll have to change this in your Kubernetes config files.
+
+#### Database Migrations
+
+The last step is to run the Fleet database migrations on your new MySQL server. To do this, run the following:
+
+```
+kubectl create -f ./examples/kubernetes/fleet-migrations.yml
```
groob: The migration job and the fleet deployment must share the same container image source (otherwise what migrations are you running?) The notes need to mention the image in the job. 

Also, another thing worth noting is that `replace -f` would delete and create the new job in a single command.

docs/infrastructure/fleet-on-kubernetes.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Setting Up and Installing Fleet
+
+### Create Server Secrets
+
+It should be noted that by default Kubernetes stores secret data in plaintext in etcd. Using an alternative secret storage mechanism is outside the scope of this tutorial, but let this serve as a reminder to secure the storage of your secrets.
+
+#### TLS Certificate & Key
+
+To generate a TLS key and a CSR (certificate signing request), you can run the following:
```
groob: Recommend `lego` with the DNS challenge here. https://github.com/xenolf/lego#cli-example Free and easy to potbain certs.

docs/infrastructure/fleet-on-kubernetes.md
```diff
@@ -0,0 +1,132 @@
...
+
+```
+openssl req -new -newkey rsa:2048 -nodes -keyout tls.key -out tls.csr
+```
+
+Now you'll have to give this CSR to a Certificate Authority, and they will give you a file called `tls.crt`. We will then have to add the key and certificate as Kubernetes secrets.
+
+```
+kubectl create secret generic fleet-tls --from-file=./tls.key --from-file=./tls.crt
```
groob: `kubectl create secret tls ` instead of generic. TLS is a special secret type.

examples/kubernetes/fleet-deployment.yml
```diff
@@ -0,0 +1,56 @@
...
+      labels:
+        app: fleet-webserver
+    spec:
+      volumes:
+        - name: fleet-tls
+          secret:
+            secretName: fleet-tls
+      containers:
+      - name: fleet-webserver
+        image: kolide/fleet:latest
```
groob: `latest` is not great in k8s, because applying the deployment again won’t pull latest. k8s versioning of imagesis done through explicit version numbers. 

Make a note of that in the docs if you want to keep latest in the docs.

examples/kubernetes/fleet-migrations.yml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: fleet-prepare-db
+spec:
+  template:
+    metadata:
+      name: fleet-prepare-db
+    spec:
+      containers:
+      - name: fleet
+        image: kolide/fleet:latest
```
groob: See note about latest.

examples/kubernetes/fleet-deployment.yml
groob: Reference https://kubernetes.io/docs/concepts/configuration/overview/#container-images

docs/infrastructure/fleet-on-kubernetes.md
groob: Handy command I use:

```
GCE_PROJECT="kolide" GCE_DOMAIN="acme-co" \
    lego --email="victor@acme.co" \
    -x "http-01" \
    -x "tls-sni-01" \
    --domains="fleet.acme.co" \
    --dns="gcloud" --accept-tos run
```

The AWS one is very similar except for the env vars. There's also a handy k8s operator that manages certs for you. https://github.com/PalmStoneGames/kube-cert-manager

Worth mentioning.

groob:
One thing that k8s misses is the log output. You could create persistent disks for fleet, but that's not great. 

My PR for oklog would be an example of how to save logs to k8s, otherwise they just kind of die with the pod.

PR #1627, opened 2017-11-16
Each file resource should map to a Fleet data model object

marpaia:
@zwass and I went through the file format and found some issues that we needed to address. This PR changes the way that decorators and labels interact with queries.


groob:
What happens if you delete a query but it's available in 5 different places (pack, labels, decorators etc).
Are you told you can't delete it, or are those resources gone with the query?

marpaia:
I think that the most reliable behavior in that case is that the server should fail to delete the query. `fleetctl` should exit with a non-zero exit code.

groob:
That sounds reasonable to me. It would also be nice to list where the query is used. 

Another question we need to resolve? what happens if the same query is referenced twice in a pack, once as a snapshot, once as differential. Right now that causes issues within osquery, but the pack format allows it. 

`fleetctl` could fail to add it with a warning, or we could silently add `differential_` to the query names when they're part of a pack.

marpaia:
Each query in a pack requires you to define `name` and `query`. In this case, `name` is the name of the query in the pack and `query` is the name of the query defined as an `OsqueryQuery` resource. I would think that the same `query` can be used twice in one pack as long as the `name` is different.

groob:
That could work, but it's not specified in the current diff.

```
apiVersion: k8s.kolide.com/v1alpha1
kind: OsqueryPack
spec:
  name: osquery_monitoring
  targets:
    labels:
      - all_hosts
  queries:
    - name: osquery_version
      interval: 7200
      snapshot: true
    - name: osquery_schedule
      interval: 7200
      removed: false
    - name: osquery_events
      interval: 86400
      removed: false
    - name: oquery_info
      interval: 600
      removed: false

```

marpaia:
There are a few places where you have to define `name` and `query`. I wonder if we can have one default to the other... Like if you don't name a `query`, then it uses the `name` of the query.

marpaia:
Consider:

```
apiVersion: k8s.kolide.com/v1alpha1
kind: OsqueryPack
spec:
  name: osquery_monitoring
  targets:
    labels:
      - all_hosts
  queries:
    - query: osquery_version
      name: osquery_version_snapshot
      interval: 7200
      snapshot: true
    - query: osquery_version
      name: osquery_version_differential
      interval: 7200
    - query: osquery_schedule
      interval: 7200
      removed: false
    - query: osquery_events
      interval: 86400
      removed: false
    - query: oquery_info
      interval: 600
      removed: false
```

groob:
As long as we have a validation layer to detect duplicates, I like this.

PR #1628, opened 2017-11-16
Fixed leaking http client

### LINKED ISSUE #1587: SAML client creates too many http connections
https://github.com/kolide/fleet/blob/5ad562a952f5e69777b0ae88a7e9ccbe49d15011/server/sso/settings.go#L74-L79

Instead of reusing the client this creates a new one with every request which will cause the server to run out of file descriptors very fast.

murphybytes:
Closes #1587.  The `GetMetadata` method now reuses `http.Client`

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -68,7 +68,7 @@
 
 [[projects]]
   name = "github.com/go-kit/kit"
-  packages = ["endpoint","log","log/level","metrics","metrics/internal/lv","metrics/prometheus","transport/grpc","transport/http"]
+  packages = ["endpoint","log","metrics","metrics/internal/lv","metrics/prometheus","transport/grpc","transport/http"]
   revision = "4dc7be5d2d12881735283bcab7352178e190fc71"
   version = "v0.6.0"
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -68,7 +68,7 @@
 
 [[projects]]
   name = "github.com/go-kit/kit"
-  packages = ["endpoint","log","metrics","metrics/internal/lv","metrics/prometheus","transport/grpc","transport/http"]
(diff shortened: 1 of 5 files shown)


server/service/service.go
```diff
@@ -14,7 +14,9 @@ import (
 	"github.com/kolide/fleet/server/kolide"
 	"github.com/kolide/fleet/server/logwriter"
 	"github.com/kolide/fleet/server/sso"
-	lumberjack "gopkg.in/natefinch/lumberjack.v2"
+	"gopkg.in/natefinch/lumberjack.v2"
+	"net/http"
```
groob: stdlib imports should go at the top.

server/service/service.go
```diff
@@ -42,6 +44,9 @@ func NewService(ds kolide.Datastore, resultStore kolide.QueryResultStore,
 		osqueryResultLogWriter: resultWriter,
 		mailService:            mailService,
 		ssoSessionStore:        sso,
+		metaDataClient: http.Client{
```
groob: usually http.Client is a pointer. not sure if it makes a huge difference, but `http.DefaultClient` is also an `&http.Client{}`

Gopkg.lock
```diff
@@ -68,7 +68,7 @@
 
 [[projects]]
   name = "github.com/go-kit/kit"
-  packages = ["endpoint","log","metrics","metrics/internal/lv","metrics/prometheus","transport/grpc","transport/http"]
+  packages = ["endpoint","log","log/level","metrics","metrics/internal/lv","metrics/prometheus","transport/grpc","transport/http"]
```
groob: I don't see this new package in the diff. Should Gopkg.lock changes be part of the commit?

marpaia:
Thanks so much for taking this on @murphybytes!

PR #1636, opened 2017-11-19
use json.RawMessage for result logs

### LINKED ISSUE #1615: osquery logger fails to flush logs to file

### LINKED ISSUE #1632: Support batch mode event logging.
By default Kolide supports event mode logging. If you enable batch mode https://github.com/facebook/osquery/blob/1cd4ed949f9077c66ecf374bd63d2c6a0a0e6e7b/docs/wiki/deployment/logging.md#batch-format then kolide is unable to parse the JSON into anything useful.

Setting `log_result_events` to true makes it work (the default).

the batch events look like:

```JSON
{
  "diffResults": {
    "added": [
      {
        "name": "osqueryd",
        "path": "/usr/local/bin/osqueryd",
        "pid": "97830"
      }
    ],
    "removed": [
      {
        "name": "osqueryd",
        "path": "/usr/local/bin/osqueryd",
        "pid": "97650"
      }
    ]
  },
  "name": "processes",
  "hostname": "hostname.local",
  "calendarTime": "Tue Sep 30 17:37:30 2014",
  "unixTime": "1412123850",
  "epoch": "314159265",
  "counter": "1"
}
```

as opposed to say the default (sorry, different query, but all I had laying around)

```JSON
{
    "action": "added",
    "calendarTime": "Thu Nov 16 21:51:09 2017 UTC",
    "columns": {
        "class": "",
        "model": "IOUSBHostDevice",
        "model_id": "8140",
        "protocol": "",
        "removable": "1",
        "serial": "0",
        "subclass": "",
        "usb_address": "8",
        "usb_port": "4",
        "vendor": "",
        "vendor_id": "0451",
        "version": "1.0"
    },
  "hostname": "hostname.local",
  "calendarTime": "Tue Sep 30 17:37:30 2014",
  "unixTime": "1412123850",
  "epoch": "314159265"
}
```

groob:
Initially fleet decoded the incoming JSON sent to the log endpoint.
Then the log event would be written to a log writer by calling json.Encoder{}.Encode.
Re-encoding logs is lossy; whenever a new field is sent by osqueryd we don't keep up with them.
Instead of caring about the content of the OsqueryResultLog, fleet will now write all log results
exactly as sent to the server by osqueryd.

Closes #1632
Closes #1615

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Kolide Fleet 1.0.6 (TBD, 2017)
+
+* Fixed bugs that caused logs to sometimes be ommited from the logwriter. (#1636, #1617)
+
 * Fixed a bug where SAML client would create too many HTTP connections. (#1587)
 
 * Fixed bug in which default query was run instead of entered query. (#1611)
--- a/server/kolide/osquery.go
+++ b/server/kolide/osquery.go
@@ -2,6 +2,7 @@ package kolide
 
 import (
 	"context"
+	"encoding/json"
 )
 
 type OsqueryService interface {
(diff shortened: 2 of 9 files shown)


groob:
@marpaia I added the batch example from #1632 as well as a more nonsensical log line which is not an actual osquery log, but is valid JSON.

marpaia:
:ship:

PR #1637, opened 2017-11-19
update deps

groob:
brings deps up to date for upcoming release.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -93,3 +93,8 @@
 
 [[constraint]]
   name = "gopkg.in/yaml.v2"
+
+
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -93,3 +93,8 @@
 
 [[constraint]]
   name = "gopkg.in/yaml.v2"
+
+
+[[constraint]]
+  name = "github.com/kolide/launcher"
+  branch = "master"
```
groob, line 100: Enforcing master for now because both projects are moving in sync. There's a fix in master for encoding log types which is not in the current launcher release.

PR #1639, opened 2017-11-19
decouple the osquery tls service from the ui service

### LINKED ISSUE #1619: move osquery TLS service into its own package
We now have a `launcher` package which implements the launcher API. It would also make sense to isolate the osquery TLS handlers into a separate package instead of bundling it with all the UI/api routes.

groob:
Closes #1619

WIP. Some of the tests have to be redone as they're too coupled to inmem.


groob:
I'm going to close this and come back to it once we refactor around fleetctl

PR #1640, opened 2017-11-19
Match release executable name to install commands

ghexp:
Current fleet_latest.zip linux executable name is fleet_linux_amd64, which doesn't match sudo cp


groob:
Thanks for the fix!

PR #1649, opened 2017-11-21
Transition osquery options interfaces for compatibility with fleetctl

zwass:
- Refinements to options yaml definition
- Datastore and service implementations
- Migration to bring existing options into new table format


zwass:
Looking for feedback on the approach used here before I continue to dive into this.

docs/cli/README.md
```diff
@@ -113,39 +113,62 @@ apiVersion: k8s.kolide.com/v1alpha1
...
-    logger_tls_period: 10
+    options:
+      distributed_interval: 3
+      distributed_tls_max_attempts: 3
+      logger_plugin: tls
+      logger_tls_endpoint: /api/v1/osquery/log
+      logger_tls_period: 10
   overrides:
-    # Note configs in overrides take precedence over base configs
+    # Note configs in overrides take precedence over the default config defined
```
groob, line 123: This is fine but I thought we were not going to use a hierarchical approach?

docs/cli/README.md
```diff
@@ -113,39 +113,62 @@ apiVersion: k8s.kolide.com/v1alpha1
 kind: OsqueryOptions
 spec:
   config:
-    distributed_interval: 3
-    distributed_tls_max_attempts: 3
-    logger_plugin: tls
-    logger_tls_endpoint: /api/v1/osquery/log
-    logger_tls_period: 10
+    options:
```
groob, line 116: Why are `options` nested? Anything else that could be used in config which will not be an option?

server/datastore/mysql/osquery_options.go
```diff
@@ -0,0 +1,125 @@
...
+	for _, row := range rows {
+		switch row.OverrideType {
+		case kolide.OptionOverrideTypeDefault:
+			spec.Config = json.RawMessage(row.Options)
+
+		case kolide.OptionOverrideTypePlatform:
+			spec.Overrides.Platforms[row.OverrideIdentifier] = json.RawMessage(row.Options)
+
+		default:
+			level.Error(d.logger).Log(
```
groob: `level.Info`

server/datastore/mysql/migrations/tables/20171116163618_CreateTableOsqueryOptions.go
```diff
@@ -0,0 +1,26 @@
...
+
+func Up_20171116163618(tx *sql.Tx) error {
+	sqlStatement := "CREATE TABLE `osquery_options` (" +
+		"`id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT," +
+		"`override_type` INT(1) NOT NULL, " +
+		"`override_identifier` VARCHAR(255) NOT NULL DEFAULT ''," +
+		"`options` JSON NOT NULL," +
+		"PRIMARY KEY (`id`)" +
+		") ENGINE=InnoDB DEFAULT CHARSET=utf8;"
+	if _, err := tx.Exec(sqlStatement); err != nil {
```
groob: could use `return errors.Wrap(err, ...) `and drop the conditional usage here.

Gopkg.lock
```diff
@@ -66,6 +66,12 @@
   revision = "47dc60e71eed504e3ef8e77ee3c6fe720f3be57f"
   version = "v1.3.0"
 
+[[projects]]
+  name = "github.com/ghodss/yaml"
```
groob, line 70: Ah just noticed. Curious why this yaml package specifically?

groob: ahh I'm seeing the doc. very cool!

zwass: I was a bit skeptical about this library, but then I saw that kubernetes is using it: https://github.com/kubernetes/kubernetes/issues/2600

docs/cli/README.md
zwass: Yes, other top-level keys are related to FIM and Yara. See the darwin override below for example.

docs/cli/README.md
zwass: Looks like I need to try to make this documentation more clear. If there is a matching override for a host, that host will receive _only_ the override config.

zwass:
> why can't config just be a single field in the db?

It can (especially now when there's only platform overrides), but I think this better prepares us for host or label based overrides in the future.

> Why are overrides hierarchical rather than a full replace?

As discussed in chat, overrides are a full replace. The language about "hierarchical" was intended to imply different precedence of various overrides (currently only platform, but in the future host would be higher precedence).

server/kolide/osquery_options.go
```diff
@@ -0,0 +1,41 @@
+package kolide
+
+import "encoding/json"
+
+type OsqueryOptionsStore interface {
+	ApplyOptions(options *OptionsSpec) error
+	GetOptions() (*OptionsSpec, error)
+	OptionsForHost(host *Host) (json.RawMessage, error)
```
zwass: The idea of `json.RawMessage` here is to support any osquery configuration without explicitly having to mention it in an OsqueryConfig object. This might be cleaner and achieve the same effect if `json.RawMessage` were replaced with `map[string]interface{}`.

We would still probably store the options as JSON in the DB, but the interface exposed by the datastore could be easier to work with.

server/datastore/mysql/osquery_options.go
```diff
@@ -0,0 +1,132 @@
...
+	err := d.db.Get(
+		&row, sql,
+		kolide.OptionOverrideTypeDefault,
+		kolide.OptionOverrideTypePlatform, host.Platform,
+		// Order of the following arguments defines precedence of
+		// overrides.
+		kolide.OptionOverrideTypePlatform, kolide.OptionOverrideTypeDefault,
+	)
+	if err != nil {
+		return nil, errors.Wrapf(err, "retrieving osquery options for host %#v", host)
```
groob: instead of printing out the whole host in the error response, maybe just the `host.ID`?

server/service/service_legacy_export.go
```diff
@@ -0,0 +1,58 @@
...
+	config := configForExport{
+		Options:   options,
+		FilePaths: fimConfig.FilePaths,
+	}
+
+	confJSON, err := json.Marshal(config)
+
+	confObj := yamlObjForExport{
+		ApiVersion: "TODO",
+		Kind:       "TODO",
```
marpaia: "Options"?

server/service/service_legacy_export.go
```diff
@@ -0,0 +1,58 @@
...
+
+	config := configForExport{
+		Options:   options,
+		FilePaths: fimConfig.FilePaths,
+	}
+
+	confJSON, err := json.Marshal(config)
+
+	confObj := yamlObjForExport{
+		ApiVersion: "TODO",
```
marpaia: The docs say `k8s.kolide.com/v1alpha1` for this

server/service/service_legacy_export.go
zwass: This file is going to be killed in favor of the migration as you suggested.

zwass:
I updated this to include a migration for options rather than the export interface it previously had. I tested manually, but it could use further testing.

zwass:
We discussed this and are moving ahead with merging the migration without automated testing. Existing data will not be effected by this migration and we should be able to assist any users who have trouble with it.

PR #1652, opened 2017-11-28
Update fleet-deployment.yaml

MattMencel:
The helm deployment of the redis service creates a service called `fleet-cache-redis`.


groob:
Nice catch! Thanks for the PR.

PR #1653, opened 2017-11-28
log the remote IP of the host making a request

groob:
A redo of #1635. Depends on https://github.com/grpc/grpc-go/pull/1696 to work. If the grpc-go change is not wanted we can try another approach.

### DIFF
--- a/server/launcher/server.go
+++ b/server/launcher/server.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/go-kit/kit/log"
+	kithttp "github.com/go-kit/kit/transport/http"
 	launcher "github.com/kolide/launcher/service"
 	grpc "google.golang.org/grpc"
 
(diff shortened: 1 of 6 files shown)


groob:
With the deps changes this should be good to 🚢 .

marpaia:
@groob did you verify that this works as intended with a launcher connected over both TLS and gRPC?

groob:
Verified with `./build/launcher -root_directory /var/groob -hostname localhost:8080 -enroll_secret_path=/Users/victor/kolide/fleet_local_test/enroll_secret -insecure -debug` and saw

```
ts=2017-11-30T23:29:33.247981Z component=service method=SubmitDistributedQueryResults ip_addr=[::1]:55836 err=null took=4.842613ms
```

previously it was a panic.

server/launcher/server.go
```diff
@@ -46,7 +47,9 @@ func New(
 func (hgprc *Handler) Handler(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.ProtoMajor == 2 && strings.Contains(r.Header.Get("Content-Type"), "application/grpc") {
-			hgprc.ServeHTTP(w, r)
+			ctx := r.Context()
+			ctx = kithttp.PopulateRequestContext(ctx, r)
```
marpaia: why not

```
ctx := kithttp.PopulateRequestContext(ctx, r.Context())
```

marpaia, line 51: why not

```
ctx := kithttp.PopulateRequestContext(r.Context(), r)
```

groob: looked cleaner to me to separate it on two lines, but I don't have a strong preference here.

marpaia:
Can you add something to the changelog as well?

PR #1655, opened 2017-12-01
add http endpoints for launcher

groob:
Depends on https://github.com/kolide/launcher/pull/209 which is a little weird to have the API endpoints defined in the client package. We'll figure it out...


groob:
Closed in favor of the Twirp PR.

PR #1657, opened 2017-12-01
In progress implementation of /api/v2 for new CLI

marpaia:
Very in progress, but putting this up for feedback before I go too deep in this direction.

### DIFF
--- a/server/api/transport_error.go
+++ b/server/api/transport_error.go
@@ -1,0 +1,153 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+)
+
+// erroer interface is implemented by response structs to encode business logic errors
+type errorer interface {
+	error() error
+}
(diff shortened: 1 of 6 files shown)


server/api/transport_error.go
```diff
@@ -0,0 +1,153 @@
...
+// same as baseError, but replaces "base" with different name.
+func namedError(name string, err string) []map[string]string {
+	return []map[string]string{map[string]string{
+		"name":   name,
+		"reason": err},
+	}
+}
+
+// encode error and status header to the client
+func encodeError(ctx context.Context, err error, w http.ResponseWriter) {
```
groob, line 38: This looks like a direct copy/paste, which is fine, but a few of the errors are no longer necessary here, notably `osqueryError`, that's only intended for osquery endpoints.

PR #1668, opened 2017-12-13
Combine service files with transports, endpoints, etc

marpaia:
This is a large refactor that we've been talking about a bit lately. As the Fleet service has grown, merging the transport, endpoints, services, middlewares, loggers, validators, etc. for each type seemed more attractive as a way to limit the number of files you needed to touch to change something.


zwass:
This is likely to cause massive merge conflicts with the `fleetctl` branch. Given that it's not changing any functionality, I'd ask that we hold off on making this change until after that has been merged back into master.

zwass:
I'm closing this, but tracking the intention to do it in https://github.com/kolide/fleet/issues/1684

Issue #1667, opened 2017-12-12
RFC: Creating CLI API requests from resources


marpaia:
## Problem

In the [`fleetctl` documentation](https://github.com/kolide/fleet/blob/master/docs/cli/README.md), we outline several file formats that define "resources". Each resource is defined in a file and version controlled (potentially). This issue discusses a proposal of how to translate these files into API requests.

## Considerations

There are two extremes that we've been considering:

### Routing in the CLI

When you execute `fleetctl apply -f ./query.yml`, the `fleetctl` tool will read the file, figure out what type of resource is defined, determine (based on static logic) what endpoint should be requested with what params, etc. 

This is nice because the requests are strongly typed and there is strong client-side validation. Unfortunately, if we add new resources, we have to make sure users have updated both the CLI and the Fleet server.

### Routing on the server

When you execute `fleetctl apply -f ./query.yml`, the `fleetctl` tool will just ship the file content to the server and the server will parse each resource in the file and determine whether or not the content is acceptable.

This is nice because the CLI doesn't need to be updated to take advantage of new resource types, but it's not so great to just have one server endpoint that accepts the text-form of `interface{}`.

## Alternative Proposal

Consider the following resource which defines an osquery query:

```
version: v2
kind: query
spec:
  name: osquery_version
  description: The version of the Launcher and Osquery process
  query: select launcher.version, osquery.version from kolide_launcher_info launcher, osquery_info osquery;
  support:
    launcher: 0.3.0
    osquery: 2.9.0
```

I propose that there is a consistent mapping of the resource metadata to request. For example, the above resource would produce a request such as:

```
POST /api/v2/kolide/query

{
  "name": "osquery_version",
  "description": "The version of the Launcher and Osquery process",
  "query: select launcher.version, osquery.version from kolide_launcher_info launcher, osquery_info osquery;",
  "support": {
    "launcher": "0.3.0",
    "osquery": "2.9.0"
  }
}
```

Basically, the body of the request would be determined by the `spec` data structure. The route would be determined by the `kind` and `version` via `/api/{version}/kolide/{kind}`. The action would determine the HTTP verb (`fleetctl delete` would be `DELETE`, `fleetctl apply` would be `POST`, etc).

This is nice because it allows us to still create a rather nice RESTful API while still allowing for enough dynamicism to allow client version to drift from server version.

PR #1670, opened 2017-12-15
Update query and pack interfaces for fleetctl

zwass:
- Add new Apply spec methods for queries and packs
- Remove now extraneous datastore/service methods
- Remove import service (unused, and had many dependencies that this breaks)
- Refactor tests as appropriate


server/kolide/queries.go
```diff
@@ -64,3 +79,63 @@ type Query struct {
 	// table in the MySQL backend.
 	Packs []Pack `json:"packs" db:"-"`
 }
+
+const (
+	QuerySpecKind = "OsqueryQuery"
+)
+
+type queryYaml struct {
```
groob: YAML is just a serialization format. IMO we should come up with a better name for the configuration data than YAML.

server/service/service_queries.go
```diff
@@ -5,8 +5,38 @@ import (
 
 	"github.com/kolide/fleet/server/contexts/viewer"
 	"github.com/kolide/fleet/server/kolide"
+	"github.com/pkg/errors"
 )
 
+func (svc service) ApplyQueryYaml(ctx context.Context, yml string) error {
+	vc, ok := viewer.FromContext(ctx)
```
groob: doesn't the `authenticatedUser` middleware handle auth already?

server/service/service_queries.go
```diff
@@ -5,8 +5,38 @@ import (
 
 	"github.com/kolide/fleet/server/contexts/viewer"
 	"github.com/kolide/fleet/server/kolide"
+	"github.com/pkg/errors"
 )
 
+func (svc service) ApplyQueryYaml(ctx context.Context, yml string) error {
```
groob: The implementation here is just serialization and could be solved in `encodeResponse`/`encodeRequest` methods no?

zwass: IMO yaml is not a transport serialization format and should not be handled in the transport (en/de)coding.

marpaia: If we follow the design enumerated in https://github.com/kolide/fleet/issues/1667, the client will never actually send the server the full yaml document.

zwass:
Pushed changes to this consistent with https://github.com/kolide/fleet/pull/1674/files

zwass:
Note this PR was massively updated to combine the refactoring of packs and queries, as these are highly intertwined.

zwass:
Also note:

We enforce referential integrity on the references from packs to queries with foreign key constraints in the DB, but our current strategy with heterogenous target types prevents us from doing the equivalent there. I think we probably want to enforce it for targets as well, but that will require changing the join table strategy for targets. Perhaps work for a later PR.

server/datastore/mysql/migrations/tables/20171219164727_UpdateScheduledQueryKey.go
```diff
@@ -0,0 +1,74 @@
...
+`
+	if _, err := tx.Exec(query); err != nil {
+		return errors.Wrap(err, "removing created_by from packs")
+	}
+
+	return nil
+}
+
+func Down_20171219164727(tx *sql.Tx) error {
+	return nil
```
marpaia, line 73: :metal:

server/kolide/queries.go
```diff
@@ -64,3 +80,63 @@ type Query struct {
...
+	Spec QuerySpec `json:"spec"`
+}
+
+type QuerySpec struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Query       string `json:"query"`
+}
+
+func LoadQueriesFromYaml(yml string) ([]*Query, error) {
```
marpaia, line 99: are these yaml functions here because you already wrote them and you're saving them for the client implementation?

server/datastore/mysql/queries.go
```diff
@@ -2,12 +2,65 @@ package mysql
...
 	"github.com/pkg/errors"
 )
 
+func (d *Datastore) ApplyQueries(authorID uint, queries []*kolide.Query) (err error) {
+	tx, err := d.db.Begin()
+	if err != nil {
+		return errors.Wrap(err, "begin ApplyQueries transaction")
+	}
+
+	defer func() {
```
marpaia, line 18: why not move this to right after the call to tx.Commit?

server/kolide/queries.go
zwass: Yes

server/datastore/mysql/queries.go
zwass: We need to call `Rollback` if any of the error cases occur between here and the call to `Commit`. This is becoming a common pattern, and I'm tempted to abstract into a function in Kolide Kit.

PR #1672, opened 2017-12-18
Fix mysql connection pooling

### LINKED ISSUE #1513: missing sql MaxOpen and MaxIdle connections.
By default, Go has an unlimited sql connection pool -- under load it will just keep growing. This causes an issue when the mysql database has a connection cap. 

See http://jmoiron.github.io/sqlx/#connectionPool

The app should set 
```
DB.SetMaxIdleConns(n int)
DB.SetMaxOpenConns(n int)
```

These values will also need to be exposed as configuration flags, as the operator will want to tweak them depending on the ability of the remote server. 
For example, with `mysqld`, the default value is 100, but CloudSQL allows up to 4000.
The number will also have to be tweaked by the operator depending on how many instances of kolide are running.

sroberts:
So I ran into an issue where 500+ hosts seems to make Kolide Fleet's web interface unusable (based on the quickstart Heroku script). 

Based on my own [lazy question](https://osquery.slack.com/archives/C1XCLA5DZ/p1513367450000172) and the help of folks like @marpaia & @groob I was pointed at issue https://github.com/kolide/fleet/issues/1513 indicating the need to cap open and idle MySQL connections. 

This is a start, but I'm working on it. Help or thoughts are welcome, but this is a current blocker for me. 

Closes #1513

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -123,8 +123,8 @@ func (man Manager) addConfigs() {
 		"MySQL TLS server name")
 	man.addConfigString("mysql.tls_config", "",
 		"MySQL TLS config value. Use skip-verify, true, false or custom key.")
-	man.addConfigInt("mysql.max_open_conns", 50, "MySQL maximum open connection handles.")
-	man.addConfigInt("mysql.max_idle_conns", 50, "MySQL maximum idle connection handles.")
+	man.addConfigInt("mysql.max_open_conns", 9, "MySQL maximum open connection handles.")
+	man.addConfigInt("mysql.max_idle_conns", 10, "MySQL maximum idle connection handles.")
 
 	// Redis
 	man.addConfigString("redis.address", "localhost:6379",
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -123,8 +123,8 @@ func (man Manager) addConfigs() {
 		"MySQL TLS server name")
 	man.addConfigString("mysql.tls_config", "",
 		"MySQL TLS config value. Use skip-verify, true, false or custom key.")
-	man.addConfigInt("mysql.max_open_conns", "50", "MySQL maximum open connection handles.")
-	man.addConfigInt("mysql.max_idle_conns", "50", "MySQL maximum idle connection handles.")
+	man.addConfigInt("mysql.max_open_conns", 50, "MySQL maximum open connection handles.")
+	man.addConfigInt("mysql.max_idle_conns", 50, "MySQL maximum idle connection handles.")
 
 	// Redis
 	man.addConfigString("redis.address", "localhost:6379",
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -25,8 +25,8 @@ type MysqlConfig struct {
 	TLSCA         string `yaml:"tls_ca"`
 	TLSServerName string `yaml:"tls_server_name"`
 	TLSConfig     string `yaml:"tls_config"` //tls=customValue in DSN
-	MaxOpenConns  string
-	MaxIdleConns  string
+	MaxOpenConns  int
+	MaxIdleConns  int
 }
 
 // RedisConfig defines configs related to Redis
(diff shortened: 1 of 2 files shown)


sroberts:
@zwass Yep, I knew someone would point me at the stuff you'd need done. It's a WIP but I'm trying to get it done quickly. I'll give those links a look and add that.

zwass:
Oh and there's one more place where you'll need to reference the config: https://github.com/kolide/fleet/blob/master/server/config/config.go#L193

sroberts:
@zwass Let me know if I wired that up correctly.

server/datastore/mysql/mysql.go
```diff
@@ -21,6 +21,7 @@ import (
 
 const (
 	defaultSelectLimit = 100000
+	mySqlThreadCount   = 100
```
zwass: This is unused now, I believe.

server/datastore/mysql/mysql.go
```diff
@@ -73,6 +74,9 @@ func New(config config.MysqlConfig, c clock.Clock, opts ...DBOption) (*Datastore
 		return nil, err
 	}
 
+	db.SetMaxIdleConns(config.MaxOpenConns)
```
zwass: These are swapped

server/config/config.go
```diff
@@ -121,6 +123,8 @@ func (man Manager) addConfigs() {
 		"MySQL TLS server name")
 	man.addConfigString("mysql.tls_config", "",
 		"MySQL TLS config value. Use skip-verify, true, false or custom key.")
+	man.addConfigString("mysql.max_open_conns", "50", "MySQL maximum open connection handles.")
```
zwass: `addConfigInt`

server/config/config.go
```diff
@@ -200,6 +204,8 @@ func (man Manager) LoadConfig() KolideConfig {
 			TLSCA:         man.getConfigString("mysql.tls_ca"),
 			TLSServerName: man.getConfigString("mysql.tls_server_name"),
 			TLSConfig:     man.getConfigString("mysql.tls_config"),
+			MaxOpenConns:  man.getConfigString("mysql.max_open_conns"),
```
zwass: `getConfigInt`

server/datastore/mysql/mysql.go
sroberts: Fixed.

server/datastore/mysql/mysql.go
sroberts: Fixed.

server/config/config.go
sroberts: Fixed. Wasn't sure about that one so I followed the pattern.

sroberts:
@zwass Alright I think I got through most of it, I'm trying to pick apart your testing framework to figure out what I need to change. I ended up just setting the default values in `config.go` to match what the testing suite seems to want:

```golang
man.addConfigInt("mysql.max_open_conns", 9, "MySQL maximum open connection handles.")
man.addConfigInt("mysql.max_idle_conns", 10, "MySQL maximum idle connection handles.")
```

I'm not sure we want these defaults to stick, but I was just going after that sweet sweet ✅ and couldn't figure out where that test suite value was coming from.

What else can I do to push this through?

zwass:
Ah @sroberts the test actually caught a bug here (I am noting test failures from https://circleci.com/gh/sroberts/fleet/9)! There was an error roundtripping the config through YAML (I know this test is hard to read, but on the plus side it caught our issue here even though we didn't explicitly include these new flags). You need to add the appropriate `yaml` tags to the struct at https://github.com/kolide/fleet/pull/1672/files#diff-c5ba45b5e0ef818c73c30250bab82c42R28. It should look like
```
	MaxOpenConns  int `yaml:"max_open_conns"`
	MaxIdleConns  int `yaml:"max_idle_conns"`
```

Please add this, then return the default values to the defaults we prefer and see if the test passes.  At that point I think we should be good to merge.

sroberts:
@zwass I think that did it!

sroberts:
@zwass Not keeping up with your build system will this be included in new Quickstart repos now or is there a build process for that first?

zwass:
@sroberts The quickstart scripts will pull the latest stable release, so this will be included when we next cut a release.

PR #1674, opened 2017-12-18
Refactor osquery options interface

zwass:
After discussion with @groob and @marpaia, we have decided that the service
methods should not be aware of any YAML/JSON definitions, and should work
directly with objects. The new pattern we will use will involve converting YAML
to JSON at the client, and then sending the JSON which will be decoded using
the familiar go-kit mechanisms before being passed to the service methods.


server/service/service_osquery_options.go
```diff
@@ -1,41 +1,20 @@
 package service
 
 import (
-	"github.com/ghodss/yaml"
```
marpaia, line 4: Is this no longer being depended on at all? If so, we can probably re-run `dep` to remove it from the lockfile.

zwass: It's still used in the test (to verify the correct JSON tags) and I think it should remain for use in the client.

PR #1680, opened 2017-12-20
Fix missing removed and shard keys in query packs

zwass:
Include the appropriate values for `removed` and `shard` when generating config
to return to osqueryd.

### DIFF
--- a/server/service/service_osquery.go
+++ b/server/service/service_osquery.go
@@ -158,6 +158,12 @@ func (svc service) GetClientConfig(ctx context.Context) (map[string]interface{},
 				Interval: query.Interval,
 				Platform: query.Platform,
 				Version:  query.Version,
+				Removed:  query.Removed,
+				Shard:    query.Shard,
+			}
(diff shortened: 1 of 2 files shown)


zwass:
Note this is an outstanding bug in `master`, but I detected it and it was easier to test in the `fleetctl` branch. We can pick the fix into `master`, but it will cause conflicts on merging `fleetctl` if we try to update the test in `master`.

marpaia:
@zwass so you're saying that if we merge this fix into `master`, there will be merge conflicts when we merge `fleetctl` into `master` because why? Restructuring to `service_osquery_test.go` in `fleetctl`?

zwass:
Yes. Major changes to the test in `fleetctl` would mean duplicated work that we would throw away almost immediately. We could totally merge the fix into master without updating the test there.

marpaia:
I think merging the fix into master without the test is a fine compromise if the bug is severe enough.

PR #1693, opened 2018-01-30
WIP: Log Forwarding

thorduri:
Hi,

This PR starts adding support for optional log forwarding of osquery status/result logs,
and rather then present a bushwack of a PR with substantial refactoring I though it best
to present a working (I'm deploying this) example for discussion.

A rough laundry list of issues to resolve:

- How to configure and support forwarding endpoints

Adding a `type` and then interface implementations would probably suffice, e.g.
```
osquery:
  status_forwarding:
    type: http
    destination: ...
    authorization: ...
---
osquery:
  status_forwarding:
    type: sumo
    destination: ...
---
osquery:
  status_forwarding:
    type: stackdriver
    project_id: ...
````

This would neatly help with dealing the differences between the endpoints, i.e. maximum
body size limits, body layout, response status code for rate-limiting etc.

- Support more then one forwarder at a time (e.g. multiple result log forwarders)
Personally, I think this has limited use and would just add complexity.

- Refactoring

I'm currently investigating if the `logwriter` package can be repurposed to own the logs
and forwarding (being renamed into `logs` or similar), and build the interface implementations
there. 

One thing to note, is that currently logs are written to disk in the endpoint handler  for `/api/v1/osquery/log`, I'm unsure if this is the way to go forward when dealing with forwarders
as they might need to retry at a later time, so refactoring that to submit to a `logs` channel
and handling later might be required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,8 @@ deps:
 	go get -u \
 		github.com/jteeuwen/go-bindata/... \
 		github.com/golang/dep/cmd/dep \
-		github.com/groob/mockimpl
+		github.com/groob/mockimpl \
+		cloud.google.com/go/logging
 	dep ensure -vendor-only
 
 distclean:
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -70,11 +70,34 @@ type SessionConfig struct {
 	Duration time.Duration
 }
 
+type OsqueryLogFile struct {
+	Path              string `yaml:"path"`
+	EnableLogRotation bool   `yaml:"enable_log_rotation"`
+}
+
+type OsqueryLogSumo struct {
+	Collector string `yaml:"collector"`
+}
+
+type OsqueryLogStackDriver struct {
+	ProjectID              string `yaml:"project_id"`
+	ApplicationCredentials string `yaml:"application_credentials"`
+}
+
+// OsqueryLog defines config related to osquery status/result log writing
+type OsqueryLog struct {
+	File        OsqueryLogFile        `yaml:"file"`
+	Sumo        OsqueryLogSumo        `yaml:"sumo"`
+	StackDriver OsqueryLogStackDriver `yaml:"stackdriver"`
+}
+
 // OsqueryConfig defines configs related to osquery
 type OsqueryConfig struct {
 	NodeKeySize         int           `yaml:"node_key_size"`
 	StatusLogFile       string        `yaml:"status_log_file"`
+	StatusLog           OsqueryLog    `yaml:"status_log"`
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by groob** [comment]
@thorduri Before you go too deep into this PR, I wanted to discuss strategy for how to implement a pluggable forwarder with you and possibly @zwass and @marpaia.

My view is that we should not be creating a "forwarder" that is separate from the log writer we have, but rather have different classes of writers (including a multiwriter?) and you use the configuration to swap one out with the other. 

The only requirements for the logWriters are that they implement the `io.Writer` interface
```
	osqueryStatusLogWriter io.Writer
	osqueryResultLogWriter io.Writer
``` 

And this can be achieved for almost any remote logging API.

Let us know your thoughts, or if I need to further clarify this.

**Comment by groob** [comment]
If you implement `io.Writer`, than the multiWriter is automatically supported through stdlib https://golang.org/pkg/io/#MultiWriter

**Comment by thorduri** [comment]
First, apologies for this bushwhack of an update. What am I'm after with this update
is if this is heading into the weeds or if a polished version of this might be acceptable.

@groob Using `io.MultiWriter` won't cut it, for a couple of reasons:
- The resulting writer is a just a ,,`io.Writer`'' (Can't type-check into `Flush`)
- First error halts the pipe

Couple of notes:
- The reason for the awful config code (can be refactored a wee bit) is to keep supporting environment variables for all of it.
- The error reporting out of the writers isn't the nicest, but does the trick. The main `Log` will never report errors unless all writers fail (but will log it).
- The dependency on `Flush` for e.g. the Sumo writer isn't up to par I think
- Unsure about keeping `context` in the `writers` interface
- `logwriter.Log` isn't pretty.
- All of the work still happens in the `SubmitLogs` endpoint handler

Config example:
```
osquery:
  ...
  status_log:
    file:
      path: /tmp/status_log_file_path
      enable_log_rotation: false
    sumo:
      collector: ...
    stackdriver:
      project_id: ...
      application_credentials: ...
  result_log:
    file:
      path: /tmp/result_log_file_path
      enable_log_rotation: false
    sumo:
      collector: ...
    stackdriver
      project_id: ...
      application_credentials: ...
```

groob:
Can you expand on this a bit more? 

> The resulting writer is a just a ,,io.Writer'' (Can't type-check into Flush)

thorduri:
> Can you expand on this a bit more?

>> The resulting writer is a just a ,,io.Writer'' (Can't type-check into Flush) 

The resulting type of an `io.MultiWriter` is [`multiWriter`](https://golang.org/src/io/multi.go),
so it's not easy to break into the individual writers to call `Flush` (I think? Wasn't obvious to me).

PR #1699, opened 2018-02-02
Check if the jwtToken is valid in authViewer.

### LINKED ISSUE #1514: JWT auth middleware does not check if token is Valid after Parse
After Parse, the JWT library [recommends](https://godoc.org/github.com/dgrijalva/jwt-go#example-Parse--Hmac) calling the Verify method. 
We don't currently do this. https://github.com/kolide/kolide/blob/e9b07fd1f04504f6240e6f1312beb87521331b5d/server/service/endpoint_middleware.go#L117-L122

Fortunately we're not vulnerable because we enforce a signing method (and Parse will return an error with a bad token). We're also using a session token inside the claim. 
Still, this oversight should be addressed.

thorduri:
Closes #1514

### DIFF
--- a/server/service/endpoint_middleware.go
+++ b/server/service/endpoint_middleware.go
@@ -95,6 +95,9 @@ func authViewer(ctx context.Context, jwtKey string, bearerToken token.Token, svc
 	if err != nil {
 		return nil, authError{reason: err.Error()}
 	}
+	if jwtToken.Valid != true {
+		return nil, authError{reason: "invalid jwt token"}
+	}
 	claims, ok := jwtToken.Claims.(jwt.MapClaims)
 	if !ok {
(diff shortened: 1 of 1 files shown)


server/service/endpoint_middleware.go
```diff
@@ -95,6 +95,9 @@ func authViewer(ctx context.Context, jwtKey string, bearerToken token.Token, svc
 	if err != nil {
 		return nil, authError{reason: err.Error()}
 	}
+	if jwtToken.Valid != true {
```
zwass, line 98: prefer `if !jwtToken.Valid {`

zwass:
@thorduri Please sign the CLA if you have a chance.

Issue #1676, opened 2017-12-19
Hosts View IP/MAC decorator missing values for Windows Hosts


monkykap:
### What version of `fleet` are you using (`fleet version --full`)?
fleet:1.0.5

### What operating system are you using?
Windows 10 enterprise 64 bit

### What did you do?
Hi, I am noticing that enrolled Windows Hosts (so far only tested )  are missing IP/MAC information next to their respective icons in the Hosts View (both list and grid). Not sure if interface name ambiguity is a cause or not.

### What did you expect to see?
IP/MAC info in decorator

### What did you see instead?
\-\-

`<span class="host-details__host-content host-details__host-content--mono">--</span>`

monkykap:
Image of issue:

![image](https://user-images.githubusercontent.com/4961844/35367765-fc73350e-0134-11e8-99fb-e742c11b2267.png)

zwass:
If anyone wants to expedite adding this info for windows hosts, you could help out by letting us know what queries need to be run to get the results.

The existing implementation is https://github.com/kolide/fleet/blob/master/server/service/service_osquery.go#L280

monkykap:
so just jotting down observations, windows does not populate the same columns with valid data as other platforms. So the query would need to be based on type (preferring DHCP), has ipv4 mask or !ipv6 mask, ordered by lower metric, and then if still a tie, ordered by interface name (text numbers).

Correction: If only query windows hosts there are many additional columns:  

connection_status | description | dhcp_enabled | dhcp_lease_expires | dhcp_lease_obtained | dhcp_server | dns_domain | dns_domain_suffix_search_order | dns_host_name | dns_server_search_order | enabled
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

as well as friendly_name

monkykap:
okay after playing around with the data more here is some proposed logic:

1. filter out any results with ipv6 masks
2. filter out any results that start with ipv4 address of 127 or 169
3. order by connection_id (which should prioritize names such as local area or ethernet, over wireless,= and vmware/virtualbox, optionally you can first order by type (dhcp will come before manual)

i think this is the right query:

select * from interface_details id join interface_addresses ia on ia.interface = id.interface where mask not like "f%" and address not like "127%" and address not like "169%" order by type,connection_id

Edit: the order by needs work...

monkykap:
Okay here's the best IMO:
[...]

PR #1700, opened 2018-02-02
improved interface querying/populating for hosts

deeso:
pseudo-fix for #1676  - I cleaned up the query interface information query to make it specific, since Windows adds some additional fields.  I also removed localhost from candidates of likely primary network interfaces when identifying the best primary interface candidate.  Also, set -1 values when fields like Error, IBytes, OBytes, etc. are unparseable.  This probably should not fail the the entire process of populating NIC data. 

More investigation/development is needed to help support hosts with multiple interfaces.  For example, when dealing with NFS Linux servers, VM Servers, NIDS, etc.

### DIFF
--- a/server/service/service_osquery.go
+++ b/server/service/service_osquery.go
@@ -277,9 +277,12 @@ var detailQueries = map[string]struct {
 	IngestFunc func(logger log.Logger, host *kolide.Host, rows []map[string]string) error
 }{
 	"network_interface": {
-		Query: `select * from interface_details id join interface_addresses ia
-                        on ia.interface = id.interface where broadcast != ""
-                        order by (ibytes + obytes) desc`,
+		Query: `select ia.interface, address, mask, broadcast, point_to_point,
+                               id.interface, mac, id.type, mtu, metric, ipackets, opackets,
+                               ibytes, obytes, ierrors, oerrors, idrops, odrops, last_change
+                        from interface_details id join interface_addresses ia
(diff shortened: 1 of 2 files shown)


deeso:
@zwass I don't have any Apple products in my (small) infrastructure right now, but it works with Windows and Ubuntu Linux 16.04.

server/service/service_osquery.go
```diff
@@ -277,9 +277,12 @@ var detailQueries = map[string]struct {
 	IngestFunc func(logger log.Logger, host *kolide.Host, rows []map[string]string) error
 }{
 	"network_interface": {
-		Query: `select * from interface_details id join interface_addresses ia
-                        on ia.interface = id.interface where broadcast != ""
-                        order by (ibytes + obytes) desc`,
+		Query: `select ia.interface, address, mask, broadcast, point_to_point,
+                               id.interface, mac, id.type, mtu, metric, flags, ipackets, opackets,
```
zwass: Please remove `flags` and `collisions` as they seem to be unused and are missing from earlier versions of osquery.

deeso:
I updated the file and nixed the two columns, and then tested the changes here locally.  

No sure if you know this already, but for future reference, Microsoft offers Windows VMs for testing browser  compatibility (Win 7 - Win 10).  These VMs are handy for testing other things as well.

zwass:
Awesome, thanks for handling this!

PR #1706, opened 2018-02-08
Remove fixed height constraint from host details component.

Jackbennett:
If some elements content extends beyond the assumed height it will overflow outside of the details block.

These changes remove any height limit, allowing flexbox to even the details height per row. Removing justify centre keeps the header in place at the top for details that are smaller than the max height in other blocks on that row.

Used flex 1 to expand the hosts details list elements to the remaining space in the details component. Can make for some odd spacing but seems nicer than bunching elements at the top of the component.

Without;
image

With changes, host html manually repeated to force flexbox to reflow and differ heights;
image

### DIFF
--- a/frontend/components/hosts/HostDetails/_styles.scss
+++ b/frontend/components/hosts/HostDetails/_styles.scss
@@ -3,7 +3,6 @@
   @include flex-direction(column);
   @include flex-grow(1);
   @include flex-basis(250px);
-  justify-content: center;
   background-color: $white;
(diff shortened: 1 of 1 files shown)


kyleknighted:
This is a view we plan on revamping with the next version, but for now I think your changes are appropriate and should hopefully alleviate some issues others are seeing as well. :shipit: @groob

groob:
Agreed.

Thanks for the PR @Jackbennett !

PR #1717, opened 2018-02-23
Add FileAccesses to FIM Configuration

### LINKED ISSUE #1708: Add Support for file_accesses in the FIM Configuration
Poking around in the latest Fleet, I was trying to enable FIM. I thought I enabled it, but I wasn't seeing file accesses. I requested the document I posted, and it's apparent that the current implementation silently ignores all `file_accesses` in the patch request.

For reference, the original patch request that added this functionality is here:
https://github.com/kolide/fleet/pull/1548/files

It appears that fixing this would mostly be a matter of adding properties to a few of the structures so that it accepts and serializes the necessary bits to the configuration. Happy to volunteer for the work, but I'm curious if you're aware of any sharp edges on this before I get started.

bcoverston:
This pull request resolves  #1708 
Along the way I discovered that the FIM interval was also not being persisted. It was an easy fix, and I was there so I threw that in as well.

### DIFF
--- a/server/kolide/app.go
+++ b/server/kolide/app.go
@@ -131,7 +131,6 @@ type AppConfig struct {
 	FIMInterval int `db:"fim_interval"`
 	// FimeFileAccess defines the FIMSections which will be monitored for file access events as a JSON formatted array
 	FIMFileAccesses string `db:"fim_file_accesses"`
-
 }
 
 // ModifyAppConfigRequest contains application configuration information
--- a/server/kolide/app.go
+++ b/server/kolide/app.go
@@ -129,6 +129,9 @@ type AppConfig struct {
 	EnableSSO bool `db:"enable_sso"`
 	// FIMInterval defines the interval when file integrity checks will occur
 	FIMInterval int `db:"fim_interval"`
+	// FimeFileAccess defines the FIMSections which will be monitored for file access events as a JSON formatted array
+	FIMFileAccesses string `db:"fim_file_accesses"`
+
 }
 
 // ModifyAppConfigRequest contains application configuration information
--- a/server/service/service_fim.go
+++ b/server/service/service_fim.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/kolide/fleet/server/kolide"
 	"github.com/pkg/errors"
+	"encoding/json"
 )
 
 func (svc service) GetFIM(ctx context.Context) (*kolide.FIMConfig, error) {
@@ -16,9 +17,16 @@ func (svc service) GetFIM(ctx context.Context) (*kolide.FIMConfig, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "getting fim paths")
 	}
+
+	var arr []string
+
+	_ = json.Unmarshal([]byte(config.FIMFileAccesses), &arr)
+
+
 	result := &kolide.FIMConfig{
 		Interval:  uint(config.FIMInterval),
 		FilePaths: paths,
+		FileAccesses: arr,
 	}
 	return result, nil
 }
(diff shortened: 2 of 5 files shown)


server/service/service_fim.go
```diff
@@ -32,7 +42,19 @@ func (svc service) ModifyFIM(ctx context.Context, fim kolide.FIMConfig) error {
 	if err != nil {
 		return errors.Wrap(err, "updating fim")
 	}
+
 	config.FIMInterval = int(fim.Interval)
+
+	if fim.FileAccesses != nil && len(fim.FileAccesses) > 0 {
```
zwass: No need for nil check, because `len` returns `0` for `nil` slices.

server/service/service_fim.go
```diff
@@ -32,7 +42,19 @@ func (svc service) ModifyFIM(ctx context.Context, fim kolide.FIMConfig) error {
...
+		fileAccesses, err := json.Marshal(fim.FileAccesses)
+		if err != nil {
+			return errors.Wrap(err, "Error creating fim section, fileaccesses must be formatted as an array [\"cassandra\",\"etc\",\"homes\"]")
+		}
+		config.FIMFileAccesses = string(fileAccesses)
+	} else {
```
zwass: Can delete this `else` branch as it will default to the empty string.

server/kolide/app.go
```diff
@@ -129,6 +129,8 @@ type AppConfig struct {
 	EnableSSO bool `db:"enable_sso"`
 	// FIMInterval defines the interval when file integrity checks will occur
 	FIMInterval int `db:"fim_interval"`
+	// FimeFileAccess defines the FIMSections which will be monitored for file access events as a JSON formatted array
```
zwass: typoed the name

zwass:
Glad to see you sorted this out! Thank you.

bcoverston:
Thanks for the review. I did some reading on go last night when I felt like my string validation was off. I hope you find this more to your liking :) Thanks for the review!

On the topic of interval, I'm willing to make some changes in this PR, but I'm not sure what the scope or intent should be. I suggest we do that in another ticket. The only think I'm adding here is what (I believe) was intended in the original FIM ticket that added it to the API.

bcoverston:
updated, let me know if there's anything else!

server/service/service_fim_test.go
```diff
@@ -11,11 +11,16 @@ import (
 )
 
 func TestGetFIMService(t *testing.T) {
+	fileAccessesString := "[\"etc\", \"home\", \"cassandra\"]"
+	fileAccessStringValue := []string([]string{"etc", "home", "cassandra"})
```
zwass: Nit: No need for outer `[]string()`, as you are already defining it as a `[]string` inside.

server/service/service_fim_test.go
```diff
@@ -38,18 +43,24 @@ func TestGetFIMService(t *testing.T) {
...
+	assert.Equal(t, resp.Interval, uint(fimIntervalTestValue))
+	assert.Equal(t, resp.FileAccesses, fileAccessStringValue)
 	paths, ok := resp.FilePaths["etc"]
 	require.True(t, ok)
 	assert.Len(t, paths, 2)
 }
 
 func TestUpdateFIM(t *testing.T) {
+	fileAccessesString := "[\"etc\", \"home\", \"cassandra\"]"
+	fileAccessStringValue := []string([]string{"etc", "home", "cassandra"})
```
zwass: Same nit

zwass:
Thanks Ben!

bcoverston:
ty!

Issue #1644, opened 2017-11-20
Document MySQL version requirement for fleet


groob:
Fleet doesn't work well with older version of MySQL, at least not without making adjustments to the default configuration. 
It also doesn't work with `mariadb`. 

We should ensure compatibility with these commonly used versions of mysql (esp mariadb), and we'll accept contributions if someone wants to take that on. But we also need to document more clearly exactly what the requirements are. 

Aside from numerous questions in chat and other support discussions, there's also a number of github issues:

- https://github.com/kolide/fleet/issues/1641
- https://github.com/kolide/fleet/pull/1598
- https://github.com/kolide/fleet/issues/1620
- https://github.com/kolide/fleet/issues/1643
- https://github.com/kolide/kolide-quickstart/issues/31

Fleet is tested and developed against `mysql 5.7` and works out of the box with that version.

kd8drx:
Worth noting: Fleet is also incompatible with Amazon Aurora, which is billed as MySQL compatible.

groob:
@kd8drx any details into what errors come up there?

znb:
@groob Just a heads up...there are issues with MySQL 5.7 that comes with Ubuntu. Please let me know if you'd like help with testing

groob:
@znb what kinds of issues? If you can provide a link or description I can start to test. 

Or did you mean your bug report you filed earlier?

znb:
@groob I meant the bug I filed earlier.

sunilkal:
@znb were you able to resolve FAIL 20161118212641_CreateTablePasswordResetRequests.go error?
 I'm running MySQL 5.7.20 on  Ubuntu 17.10.1 unable to get around it.

znb:
@sunilkal No...I'm using Ubuntu 16.04 and MySQL 5.7.something. Waiting for feedback from @groob

groob:
@znb @sunilkal I just provisioned a Ubuntu 16.04 server on DigitalOcean and installed mysql with https://www.digitalocean.com/community/tutorials/how-to-install-mysql-on-ubuntu-16-04 then followed the instructions on the docs page https://github.com/kolide/fleet/blob/63d60d5f6a0ec3239279250aa0a103f93d74b4ae/docs/infrastructure/fleet-on-ubuntu.md

```
root@ubuntu-1gb-nyc3-01:~# mysqld --version
[...]

PR #1720, opened 2018-03-05
Remove default timestamp to fix strict error

tpyo:
Running `fleet prepare db` on Ubuntu 16.04 with MySQL 5.7 fails for the reason @groob [pointed out here](https://github.com/kolide/fleet/issues/1644#issuecomment-347248450).  

The MySQL error is: `ERROR 1067 (42000): Invalid default value for 'expires_at'`

Removing the default on `expires_at` and using NOW() on INSERT resolves.


marpaia:
Sorry for not merging this sooner @tpyo. It's unacceptable to have let this sit for so long.

PR #1742, opened 2018-04-24
Fix distributed query status parsing for osquery 3

zwass:
A change from strings to integers broke the existing code. We now support
either for current and backwards compatibility.

### DIFF
--- a/server/kolide/osquery.go
+++ b/server/kolide/osquery.go
@@ -17,7 +17,7 @@ type OsqueryService interface {
 	// for) should be returned. Returning 0 for this will not activate the
 	// feature.
 	GetDistributedQueries(ctx context.Context) (queries map[string]string, accelerate uint, err error)
-	SubmitDistributedQueryResults(ctx context.Context, results OsqueryDistributedQueryResults, statuses map[string]string) (err error)
+	SubmitDistributedQueryResults(ctx context.Context, results OsqueryDistributedQueryResults, statuses map[string]OsqueryStatus) (err error)
 	SubmitStatusLogs(ctx context.Context, logs []json.RawMessage) (err error)
 	SubmitResultLogs(ctx context.Context, logs []json.RawMessage) (err error)
 }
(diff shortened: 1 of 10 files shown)


server/service/transport_osquery.go
```diff
@@ -73,10 +74,25 @@ func decodeSubmitDistributedQueryResultsRequest(ctx context.Context, r *http.Req
...
 
+	// Statuses were represented by strings in osquery < 3.0 and now
+	// integers in osquery > 3.0. Massage to string for compatibility with
+	// the service definition.
+	statuses := map[string]string{}
+	for query, status := range shim.Statuses {
+		switch s := status.(type) {
+		case string:
+			statuses[query] = s
+		case float64:
```
groob: is it really a float and not an int?

server/service/transport_osquery.go
```diff
@@ -73,10 +74,25 @@ func decodeSubmitDistributedQueryResultsRequest(ctx context.Context, r *http.Req
 		results[query] = queryResults
 	}
 
+	// Statuses were represented by strings in osquery < 3.0 and now
+	// integers in osquery > 3.0. Massage to string for compatibility with
+	// the service definition.
+	statuses := map[string]string{}
```
groob: should we make deeper changes and use `map[string]int` (or float64`) instead of converting the number into a string?

server/service/transport_osquery.go
zwass: JSON numbers always become `float64` when you unmarshal them into `interface{}`.

server/service/transport_osquery.go
zwass: I refactored the types (added `kolide.OsqueryStatus`). This may be worth reviewing again.

server/kolide/osquery.go
```diff
@@ -26,6 +26,15 @@ type OsqueryService interface {
 // osquery distributed query.
 type OsqueryDistributedQueryResults map[string][]map[string]string
 
+// OsqueryStatus represents osquery status codes (0 = success, nonzero =
+// failure)
+type OsqueryStatus int
+
+const (
+	// StatusOK is the success code returned by osquery
+	StatusOK OsqueryStatus = 0
```
groob, line 35: ` StatusOK OsqueryStatus = iota` ?

zwass: I'd prefer to leave it as the explicit zero since this isn't really an enum but a constant for the `0` status code returned from osquery.

PR #1743, opened 2018-04-24
Include RFC822 From header.

angrylogic:
The DMARC and DKIM email authentication systems both require the RFC822
`From` header to function.  Kolide currently only includes the configured
sender address as the SMTP Envelop From address (e.g., the MAIL FROM
command).  

This patch also includes the configured sender address in the
RFC822 email `From` header which should allow these emails to pass both
DKIM and DMARC authentication.


zwass:
Thanks! We'll be happy to merge this when you sign the CLA.

angrylogic:
CLA signed, thanks!

PR #1751, opened 2018-05-02
Add fleetctl config and auth commands

marpaia:
```
$ fleetctl config set address https://localhost:8080
[+] Set the "address" config key to "https://localhost:8080" in the "default" context

$ fleetctl config set ignore_tls true
[+] Set the "ignore_tls" config key to "true" in the "default" context

$ fleetctl setup --email mike@arpaia.co --password "abc123"
[+] Fleet setup successful and context configured!

$ cat ~/.fleet/config
contexts:
  default:
    address: https://localhost:8080
    email: mike@arpaia.co
    ignore_tls: true
    token: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzZXNzaW9uX2tleSI6IlUvdm05Vk9wSG0xUlA4SUtjQnBhb2ovWlo1TXppSEVXcFRCNFNPb2tHQnNLUFpDQXFieVpWWnpJb0UvczQzcWkyd1pHZXJOa29SNFVIQ2hNZUc0K09RPT0ifQ.rHawSN8JvD4jjWAPTYX2Ep9ZpMt3u4mSIQcu920C-_s

$ fleetctl logout
[+] Fleet logout successful and local token cleared!

$ cat ~/.fleet/config
contexts:
  default:
    address: https://localhost:8080
    email: mike@arpaia.co
    ignore_tls: true
    token: ""
```


server/service/endpoint_setup.go
```diff
@@ -7,14 +7,14 @@ import (
 	"github.com/kolide/fleet/server/kolide"
 )
 
-type setupRequest struct {
+type SetupRequest struct {
```
zwass: I feel like these types should be moved into `package kolide` if they're going to be public.

marpaia: I think I'm going to keep them private and move the client into `package service`

marpaia: But I'm not too sure about this really.

marpaia:
I think `Do` should return a `*http.Response`, but I'm not opposed to extracting some shared helpers that accept a `*http.Response` for consistent JSON parsing.

PR #1752, opened 2018-05-02
Add endpoints for apply/get of each spec type

zwass:
- Osquery Options (to be renamed to Options after current Options removed)
- Labels
- Queries
- Packs

### DIFF
--- a/server/service/handler.go
+++ b/server/service/handler.go
@@ -45,11 +45,15 @@ type KolideEndpoints struct {
 	ModifyQuery                    endpoint.Endpoint
 	DeleteQuery                    endpoint.Endpoint
 	DeleteQueries                  endpoint.Endpoint
+	ApplyQuerySpecs                endpoint.Endpoint
+	GetQuerySpecs                  endpoint.Endpoint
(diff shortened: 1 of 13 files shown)


server/service/handler.go
```diff
@@ -372,15 +404,22 @@ func attachKolideAPIRoutes(r *mux.Router, h *kolideHandlers) {
 	r.Handle("/api/v1/kolide/queries/{id}", h.ModifyQuery).Methods("PATCH").Name("modify_query")
 	r.Handle("/api/v1/kolide/queries/{id}", h.DeleteQuery).Methods("DELETE").Name("delete_query")
 	r.Handle("/api/v1/kolide/queries/delete", h.DeleteQueries).Methods("POST").Name("delete_queries")
+	r.Handle("/api/v1/kolide/queries/spec", h.ApplyQuerySpecs).Methods("POST").Name("apply_query_specs")
```
marpaia, line 391: how about `/api/v1/kolide/queries/apply`?

marpaia: that's a good idea!

marpaia: :shipit:

PR #1760, opened 2018-05-07
Implement clients for labels, packs and queries

zwass:
Go client methods for making get, set and delete requests on label, pack and
query entities.


server/service/client_labels.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/kolide/fleet/server/kolide"
+	"github.com/pkg/errors"
+)
+
+// ApplyLabelSpecs sends the list of Labels to be applied (upserted) to the
+// Fleet instance.
+func (c *Client) ApplyLabelSpecs(specs []*kolide.LabelSpec) error {
+	req := applyLabelSpecsRequest{Specs: specs}
+	response, err := c.Do("POST", "/api/v1/kolide/spec/labels", req)
```
marpaia: i thought this was `/api/v1/kolide/labels/spec`?

PR #1762, opened 2018-05-07
Working prototype of fleetctl apply

marpaia:


cmd/fleetctl/apply.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	for _, spec := range strings.Split(string(b), "---") {
+		if strings.TrimSpace(spec) == "" {
+			continue
+		}
+
+		var s specMetadata
+		if err := yaml.Unmarshal([]byte(spec), &s); err != nil {
+			return nil, errors.Wrap(err, "error unmarshaling spec")
```
zwass: I like to leave "error" out of the messages in `errors.Wrap` because we end up with `error foo....: error bar...: error baz....`. Usually we log these in a way that makes it clear an error occurred anyway.

cmd/fleetctl/apply.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/ghodss/yaml"
+	"github.com/kolide/fleet/server/kolide"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
+)
+
+type specMetadata struct {
+	Kind    string      `json:"kind"`
+	Version string      `json:"apiVersion"`
+	Spec    interface{} `json:"spec"`
```
zwass: I'd prefer to make this `json.RawMessage`, or leave it out so that it doesn't have to be parsed multiple times.

PR #1768, opened 2018-05-08
Add fleetctl get label, pack, query for individual items by name

marpaia:
This PR adds support for getting resources by name.

```
$ fleetctl get queries
no queries found

$ fleetctl apply -f ./query.yaml
[+] applied 1 queries

$ fleetctl get queries
+-----------------+--------------------------------+--------------------------------+
|      NAME       |          DESCRIPTION           |             QUERY              |
+-----------------+--------------------------------+--------------------------------+
| osquery_version | The version of the Launcher    | select launcher.version,       |
|                 | and Osquery process            | osquery.version from           |
|                 |                                | kolide_launcher_info launcher, |
|                 |                                | osquery_info osquery;          |
+-----------------+--------------------------------+--------------------------------+

$ fleetctl get query osquery_version
apiVersion: v1
kind: query
spec:
  description: The version of the Launcher and Osquery process
  name: osquery_version
  query: select launcher.version, osquery.version from kolide_launcher_info launcher,
    osquery_info osquery;
```

### DIFF
--- a/cmd/fleetctl/get.go
+++ b/cmd/fleetctl/get.go
@@ -4,10 +4,18 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/ghodss/yaml"
+	"github.com/kolide/fleet/server/kolide"
 	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 )
+
+type specGeneric struct {
+	Kind    string      `json:"kind"`
+	Version string      `json:"apiVersion"`
+	Spec    interface{} `json:"spec"`
+}
 
 func defaultTable() *tablewriter.Table {
 	table := tablewriter.NewWriter(os.Stdout)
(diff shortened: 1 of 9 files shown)


cmd/fleetctl/get.go
```diff
@@ -61,7 +69,23 @@ func getQueriesCommand() cli.Command {
...
+					Version: kolide.ApiVersion,
+					Spec:    query,
+				}
+
+				b, err := yaml.Marshal(spec)
+				if err != nil {
+					return err
+				}
+
+				fmt.Printf(string(b))
```
zwass: Does it already have a newline?

zwass: Also, let's avoid using `Printf` without a format string.

marpaia: It does already have a newline. I'll use `fmt.Print`.

PR #1772, opened 2018-05-08
Consistent permissions for ~/.fleet/config file

zwass:


cmd/fleetctl/config.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/urfave/cli"
 )
 
+const configFilePerms = 0600
```
marpaia: nit: this could be:

```
const (
  foo = "bar"
)
```

PR #1779, opened 2018-05-09
Rename ignore-tls flag to tls-skip-verify

zwass:


zwass:
My assertion is that `--insecure` will be much more intuitive to folks who are used to using that flag with `curl`.

zwass:
I don't think it's the most accurate flag for what's going on. `--ignore-tls` gives the impression that TLS is not being used. What's really happening is that we are skipping certificate verification.

marpaia:
I could see `--ignore-tls-cert` or `--tls-no-verify` then maybe?

zwass:
I like `tls-no-verify` (or `tls-skip-verify`)

zwass:
Updated to `tls-skip-verify`

PR #1781, opened 2018-05-11
Commit bindata.go for library usage

marpaia:
Using Fleet as a library fails because `undefined: Asset`.


marpaia:
@zwass I don't plan on merging this, I just need the branch on origin so that I can import fleet as a library in another project. I'll close this for now and we can figure out how to move `Asset` from `package service`.

PR #1784, opened 2018-05-16
Prototype of fleetctl query command

zwass:


zwass:
This still needs work, but it's a good opportunity for folks to evaluate the strategy.

Known TODOs
- Support targeting hosts/labels by name rather than ID
- Add client support for targeting

server/service/client_live_query.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/kolide/fleet/server/kolide"
+	ws "github.com/kolide/fleet/server/websocket"
+
+	"github.com/gorilla/websocket"
+	"github.com/pkg/errors"
+)
+
+// LiveQueryResultsHandler provides access to all of the information about an
+// incoming stream of live query results.
+type LiveQueryResultsHandler struct {
```
zwass, line 19: I would like particular attention on the client exposed here.

In this implementation, errors and results are pushed to separate channels that _must both_ have readers (otherwise the reading thread will block on writing to the channel that does not have a reader). Totals and status information are not pushed over channels, but rather atomically updated so that the client user can get the up-to-date value whenever they like.

Does this seem reasonable? Other ideas on how to structure this client?

marpaia: Both channels needing readers seems like an unfortunate API requirement, but I think it would be reasonable if there was only one channel that required a reader (the results channel)? Can we represent a cancel/error check in a way that doesn't require a reader?

marpaia: I don't know if the type assertion is a real win for clarity. I think we should just merge this and iterate on the interface as we play with it in the command.

zwass:
I think the CLI interface could be improved upon, but I'd say let's merge this and iterate.

PR #1785, opened 2018-05-16
Check if the jwtToken is valid in authViewer.

### LINKED ISSUE #1514: JWT auth middleware does not check if token is Valid after Parse

thorduri:
Closes #1514 

Dup of #1699 (Have signed the CLA)

### DIFF
--- a/server/service/endpoint_middleware.go
+++ b/server/service/endpoint_middleware.go
@@ -95,6 +95,9 @@ func authViewer(ctx context.Context, jwtKey string, bearerToken token.Token, svc
 	if err != nil {
 		return nil, authError{reason: err.Error()}
 	}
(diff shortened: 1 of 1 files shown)


marpaia:
Thanks @thorduri!

PR #1786, opened 2018-05-16
ensure email and passwords are set during setup

thorduri:
Caused me a wee bit of confusion, when toying with `fleetctl setup`.

This boils down to the setup endpoint not being wrapped in the usual validation middleware,
figured this should be fine, as all other knobs can be managed after login (Name, username,
orgname etc).

### DIFF
--- a/cmd/fleetctl/setup.go
+++ b/cmd/fleetctl/setup.go
@@ -49,6 +49,12 @@ func setupCommand() cli.Command {
 				return err
 			}
 
+			if flEmail == "" {
+				return errors.Errorf("Email of the admin user to create must be provided")
+			}
+			if flPassword == "" {
+				return errors.Errorf("Password for the admin user to create must be provided")
+			}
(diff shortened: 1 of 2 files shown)


cmd/fleetctl/setup.go
```diff
@@ -49,6 +49,12 @@ func setupCommand() cli.Command {
 				return err
 			}
 
+			if flEmail == "" {
+				return fmt.Errorf("Email of the admin user to create must be provided")
```
marpaia: errors.Errorf here too :)

thorduri: Rest of `fleetctl` uses `fmt.Errorf`, and there is mixed usage in `server/` and `fleet`.

Project wide `s/fmt.Errorf/errors.Errorf` ?

marpaia: Yup, we're just progressively refactoring for now but we can totally put together a PR for a complete refactor.

PR #1788, opened 2018-05-17
server: formatting

thorduri:
My editor is pretty zealous about running `go fmt` 😁 

Any objections to adding lint targets to `circleci`  ?

It would be mess right now to add all of [golint](https://github.com/golang/lint) without a bunch of ignore handling, but `gofmt` should be there just to reduce accidental noise in diffs ?


marpaia:
I would love to get `gometalinter` going in Circle, we use it in a few other projects and it works well. Thanks for this!

Issue #1703, opened 2018-02-07
Support allowNativePasswords in DB connection


joncrain:
Currently fleet does not allow the `allowNativePasswords` DSN parameter to enable connections to Azure MySQL. Instead of passing a bunch of different params, requesting to simply expose the DSN string instead.

More information from a similar request available here: https://github.com/grafana/grafana/pull/8618

### What version of `fleet` are you using (`fleet version --full`)?
 - version 1.0.6
  branch: 	master
  revision: 	45165aa29aae52346ed1c458f19b89947d315c61
  build date: 	2017-12-04T22:50:27Z
  build user: 	marpaia
  go version: 	go1.9.2
### What operating system are you using?
Ubuntu 16.04.3 LTS
Azure MySQL
### What did you do?
fleet serve --config ./confg.yml
### What did you expect to see?
Proper connection
### What did you see instead?
"mysql":"could not connect to db: this user requires mysql native password authentication.

audibleblink:
+1

Would love to use Azure MySQL for our Fleet deployment. We're nearing 5k machines.

zwass:
Fixed by #1805

PR #1805, opened 2018-06-01
passes allowNativePasswords DSN setting

audibleblink:
Azure's MySQL implementation appears to always respond with an Authentication Method Switch Request Packet requesting mysql_native_password auth even though the client's initial Handshake Response already included the native password credentials.

Most MySQL client libraries support this but Golang's go-sql-driver/mysql requires the allowNativePasswords DSN parameter set to enable it. Without this parameter, Fleet fails to authenticate to the database with the error message "this user requires mysql native password authentication."

References:

    https://web.archive.org/web/20160814002743/http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchRequest
    https://web.archive.org/web/20160814002743/http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::HandshakeResponse
    https://github.com/go-sql-driver/mysql#user-content-allownativepasswords

### DIFF
--- a/server/datastore/mysql/mysql.go
+++ b/server/datastore/mysql/mysql.go
@@ -272,7 +272,7 @@ func registerTLS(config config.MysqlConfig) error {
 // provided configuration.
 func generateMysqlConnectionString(conf config.MysqlConfig) string {
 	dsn := fmt.Sprintf(
-		"%s:%s@(%s)/%s?charset=utf8mb4&parseTime=true&loc=UTC&clientFoundRows=true",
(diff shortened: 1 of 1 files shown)


audibleblink:
Addresses #1703

zwass:
I did some research and saw that this is the default starting in the v1.4 MySQL Driver (not yet released). Given that I feel comfortable merging this.

https://github.com/go-sql-driver/mysql/pull/644

zwass:
Thanks for sorting that out @audibleblink.

PR #1819, opened 2018-06-11
Attempt 2 at returning a serializable config

marpaia:


marpaia:
Tested this with osqueryd TLS plugins.

PR #1829, opened 2018-06-14
Fix pack and query UI issues in Fleet 2.0

### LINKED ISSUE #1810: fleet 2.0 does not allow editing packs in the GUI
Response is always `405 Method Not Allowed` unless I use fleetctl.

### LINKED ISSUE #1811: fleet 2.0 DELETE on existing query results in 404
```
{
  "message": "Resource Not Found",
  "errors": [
    {
      "name": "base",
      "reason": "queries 1 was not found in the datastore"
    }
  ]
}
```

zwass:
Replaces (and appropriately refactors) a number of endpoints that were removed long ago when we decided to kill the UI with the fleetctl release. We turned out not to do this, and now need to restore these missing endpoints.

This is not a straight up replacement of the existing code because of refactoring to the DB schemas that was also done in the migration.

Most of the replaced code was removed in #1670 and #1686.

Fixes #1811, fixes #1810


server/datastore/mysql/packs.go
```diff
@@ -288,6 +349,70 @@ func (d *Datastore) ListPacks(opt kolide.ListOptions) ([]*kolide.Pack, error) {
 	return packs, nil
 }
 
+// AddLabelToPack associates a kolide.Label with a kolide.Pack
+func (d *Datastore) AddLabelToPack(lid uint, pid uint, opts ...kolide.OptionalArg) error {
+	db := d.getTransaction(opts)
+
+	query := `
+		INSERT INTO pack_targets ( pack_id,	type, target_id )
```
marpaia: there's a weird space in this query

PR #1877, opened 2018-07-27
updated fonts and changed hosts to devices

ccoy:
This PR makes two initial updates to the Fleet UI:

Replaces Oxygen with Lato as primary font
Replaces "Hosts" with "Devices," except in Labels. That will need to be addressed in a future update.


sabrinabuckets:
@zwass That change comes from the mocks @FritzX6 created. I confirmed with him before making the change that it was something we still wanted to do. As for the change being inconsistently applied, the remaining instances were identified but potentially require a migration. I was planning to file an issue so that someone can assist me with those.

zwass:
@ccoy do you mind if I split up this PR to include the font change and then handle the copy change all at once in a separate PR?

sabrinabuckets:
@zwass before we do that, would you mind taking a look with me at what it would take to get the remaining bits of copy updated? If it's less difficult than we thought last week then I would rather get in in now.

zwass:
We could rename the label, but that would mean that folks who are currently targeting `All Hosts` with `fleetctl query` would have to target `All Devices`. Maybe we can follow up with @FritzX6 to decide whether this tradeoff is worth it?

sabrinabuckets:
Determined that this requires more effort than is currently worthwhile. I am going to close this PR and open a fresh one for just the font changes.

PR #1879, opened 2018-07-27
updates to color scheme

ccoy:
This PR makes minor changes to Fleet UI color scheme, to update to bring up to date with current Kolide branding.


zwass:
Can you please add a before/after screenshot here so we can easily tell what the changes look like?

sabrinabuckets:
after
before

PR #1887, opened 2018-08-02
adding two simple cli args for query: --quiet and --exit

jrossi:
--quiet disable all printing other then errors and results
--exit will quit and exit when 100% results are received.

### DIFF
--- a/cmd/fleetctl/query.go
+++ b/cmd/fleetctl/query.go
@@ -47,13 +47,13 @@ func queryCommand() cli.Command {
 				Name:        "quiet",
 				EnvVar:      "QUIET",
 				Destination: &flQuiet,
-				Usage:       "Quiet and don't show anything other the results",
+				Usage:       "Only print results (no status information)",
 			},
 			cli.BoolFlag{
 				Name:        "exit",
 				EnvVar:      "EXIT",
 				Destination: &flExit,
-				Usage:       "Only print results (no status information)",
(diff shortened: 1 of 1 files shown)


cmd/fleetctl/query.go
```diff
@@ -43,6 +43,18 @@ func queryCommand() cli.Command {
 				Destination: &flLabels,
 				Usage:       "Comma separated label names to target",
 			},
+			cli.BoolFlag{
+				Name:        "quiet",
+				EnvVar:      "QUIET",
+				Destination: &flQuiet,
+				Usage:       "Quiet and don't show anything other the results",
```
zwass: Let's change this to "Only print results (no status information)".

cmd/fleetctl/query.go
```diff
@@ -116,7 +130,19 @@ func queryCommand() cli.Command {
 							percentOnline = 100 * float64(responded) / float64(online)
 						}
 					}
-					s.Suffix = fmt.Sprintf("  %.f%% responded (%.f%% online) | %d/%d targeted hosts (%d/%d online)", percentTotal, percentOnline, responded, total, responded, online)
+
+					if percentTotal >= 100 && flExit {
```
zwass: I'd prefer to compare the actual values rather than the percentages.

jrossi: Updated

PR #1888, opened 2018-08-02
Fleetctl query improvements

thorduri:
There is no reason to keep looping on results when all hosts in the fleet have responded, so exit
when that happens. 

An option to exit when all online hosts have responded might be useful, but isn't included in this
PR.

Pausing the spinner when outputting prevents "ugly" output on the terminal when stderr isn't
redirected to /dev/null.

### DIFF
--- a/cmd/fleetctl/query.go
+++ b/cmd/fleetctl/query.go
@@ -106,9 +106,11 @@ func queryCommand() cli.Command {
 				select {
 				case hostResult := <-res.Results():
 					out := resultOutput{hostResult.Host.HostName, hostResult.Rows}
+					s.Stop()
 					if err := json.NewEncoder(os.Stdout).Encode(out); err != nil {
 						fmt.Fprintf(os.Stderr, "Error writing output: %s\n", err)
 					}
+					s.Start()
 
 				case err := <-res.Errors():
 					fmt.Fprintf(os.Stderr, "Error talking to server: %s\n", err.Error())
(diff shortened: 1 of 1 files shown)


thorduri:
@zwass Thanks.

#1887 accomplishes the same as this PR but behind a knob (`--exit`), I think it should be the default
behaviour, and shouldn't require a knob (this PR). 

As for my final suggestion, that would be something like `--exit-on-online-responded`, and a check
like `if knob && responded == online`. Which is AFAICT not what #1887 does.

zwass:
Ah yeah I agree it should be the default behavior. I thought #1887 implemented exiting when all _online_ hosts had responded. I think we'll make that the behavior in #1887 and use your default behavior.

cmd/fleetctl/query.go
```diff
@@ -92,15 +92,18 @@ func queryCommand() cli.Command {
...
 						fmt.Fprintf(os.Stderr, "Error writing output: %s\n", err)
 					}
+					s.Start()
 
 				case err := <-res.Errors():
 					fmt.Fprintf(os.Stderr, "Error talking to server: %s\n", err.Error())
 
 				case <-tick.C:
 					// Print status message to stderr
+					statusfmt := " %.f%% responded (%.f%% online) | %d/%d targeted hosts (%d/%d online)"
```
zwass: Can we `fmt.Sprintf` this status string once, then output it as appropriate lower down? I'd prefer not to see all the values plugged in multiple times (on line 122 and line 125).

zwass:
Can you please rebase this now that #1887 has been merged?

thorduri:
Ping @zwass

FritzX6:
@thorduri , FYI Zach is away until Sept 7

zwass:
Thanks @thorduri!

Issue #1875, opened 2018-07-26
External Data Sinks for OSQuery Results and Status messages


jrossi:
I am about to spend to much time on planes and as PoC I am planning on adding osquery results logging directly into fleet.  This would allow the results to be processed downstream and outside of fleet. 

Normally I would just reach for kafka and move on, but after looking at the code base I think NATs https://github.com/nats-io/gnatsd might be a slightly better option.  Here is why: 

- NATs  installation is easier 
- using NATs in place of Redis for web sockets is much simpler
- code is much simpler 

With that being said ignoring kafka is crazy, but for a PoC I think simple is better.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jrossi:
#1899 was created as Proof of Concept of this.  I open the PR well before it's done to gather feedback early :)

jrossi:
I have a much better understand of how Fleet works, and how the code is put together after doing my POC in #1899.  With this I would like to proposal the following as a method for integration of external data sinks for osquery results/status.  

## Proposal

Rather then adding more server configuration that needs to be completed on hosts running fleet.  It would be simpler to add the routing of data to sinks using apiVersioned yaml specification.  

This would keep the server configuration very simple while also allowing a simple path users that upgrade.  Here is an example of what the config would look like.  
```
apiVersion: v1
kind: output
spec:
  name: windows_kafka_queue
  descriptions: All windows hosts request pro
  type: kafka 
  kafka:
     topic: fleet.windows.results
     brokers: 127.0.0.1:9018
  match:
    - type: results
    - platform: windows
---
apiVersion: v1
kind: output
spec:
  name: unix_nats_queue
  descriptions: All
  type: kafka 
  kafka:
     topic: fleet.unix.{.PlatForm}
     brokers: 127.0.0.1:9018
  match:
    - type: results
    - platform: !windows
---
apiVersion: v1
kind: output
spec:
  name: local_results
  descriptions: All
  type: file
  file:
     path: /tmp/fleet_results
     log_rotate: false
  match:
    - type: results
---
apiVersion: v1
kind: output
spec:
  name: local_status
  descriptions: All
  type: file
  file:
     path: /tmp/fleet_status
     log_rotate: false
  match:
    - type: status
```

marpaia:
I think we should go back and forth on the file format here, but I love the idea of defining logging pipeline details with the file format! Great idea, @jrossi.

zwass:
[...]

PR #1899, opened 2018-08-10
[WIP] [POC] Nats support for osquery results, status and fleet campaigns

jrossi:
This is very much a Proof of Concept and a Pull Request to get feedback.  This Pull Request changes more then one thing, and to work needed to change the structure of how fleet did some things.

### DIFF
--- a/server/kolide/health.go
+++ b/server/kolide/health.go
@@ -1,0 +1,6 @@
+package kolide
+
+
+type HealthService interface {
+	HealthCheckers() (map[string]interface{})
+}
--- a/server/kolide/health.go
+++ b/server/kolide/health.go
@@ -1,6 +1,0 @@
-package kolide
-
-
-type HealthService interface {
-	HealthCheckers() (map[string]interface{})
-}
--- a/server/kolide/queue.go
+++ b/server/kolide/queue.go
@@ -1,4 +1,10 @@
 package kolide
 
+import (
+	"context"
+	"encoding/json"
+)
+
 type QueueService interface {
+	Messages(ctx context.Context, logs []json.RawMessage) error
 }
--- a/server/kolide/queue.go
+++ b/server/kolide/queue.go
@@ -1,0 +1,4 @@
+package kolide
+
+type QueueService interface {
+}
--- a/server/kolide/queue.go
+++ b/server/kolide/queue.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 )
 
+// QueueService defines functions for sending query results over to external
+// system. It is implemented by structs in package queue.
 type QueueService interface {
 	Messages(ctx context.Context, logs []json.RawMessage) error
 }
--- a/server/queue/file.go
+++ b/server/queue/file.go
@@ -1,0 +1,73 @@
+package queue 
+
+import (
+	"os"
+	"fmt"
+	"os/signal"
+	"syscall"
+	"context"
+	"encoding/json"
+
+	kitlog "github.com/go-kit/kit/log"
+	"github.com/kolide/fleet/server/config"
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+
+type FileQueue struct {
+	logger         kitlog.Logger
+	enableRotation bool
+	path           string
+	l              *lumberjack.Logger
+}
+
+func NewFileQueue(appLogger kitlog.Logger, conf config.FileQueueConfig) (*FileQueue, error){
+	var osquerydLogger *lumberjack.Logger
+
+	if conf.EnableLogRotation {
+		osquerydLogger = &lumberjack.Logger{
+			Filename:   conf.LogFile,
+			MaxSize:    500, // megabytes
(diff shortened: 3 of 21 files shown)


zwass:
Thanks for working on this. I think the `CampaignStore` implementation for NATS is pretty uncontroversial and we should not have much trouble merging it. If you'd like to accelerate that you could open a separate PR where we just look at that implementation.

What I'd really like to see is more documentation in the code and in the pull request description about the new interfaces you are introducing and how they effect the behavior of Fleet. We would also need to come up with a plan to make the transition easy for existing Fleet users.

jrossi:
@zwass I completely agree that this needs a lot more docs, details, and what not.  Along the way some feedback on the direction and path would be helpful so as to not spend to much time working down dead ends.  I am going to add them into the pull request directly as review comment.  Might be easier with the context of the code. 

As for the Nats CampaignStore.  I don't see any value in pulling that out as redis is still required for sessions storage, but happy to do this if breaking out this part makes sense.

server/queue/file.go
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+func (fq *FileQueue) Messages(ctx context.Context, logs []json.RawMessage) error {
+	for _, log := range logs {
+		if _, err := fq.l.Write(append(log, '\n')); err != nil {
+			return fmt.Errorf("error writing status log: " + err.Error())
+		}
+	}
+	/*
```
jrossi, line 65: This needs to be corrected to be the old implementation that Flushes

server/queue/nats.go
```diff
@@ -0,0 +1,53 @@
...
+	topicTemplate *template.Template
+	conn          *nats.Conn
+}
+
+func NewNatsQueue(appLogger kitlog.Logger, conn *nats.Conn, conf config.NatsQueueConfig) (*NatsQueue, error) {
+	nq := &NatsQueue{
+		conn:      conn,
+		topicRaw:  conf.Topic,
+	}
+	t, err := template.New("nats_topic").Parse(nq.topicRaw)
```
jrossi, line 30: Does the use of text/template to give the use some choice on how to route the output make sense?  I thought this was a powerful idea and feature as you could do templates like the following: 

`fleet.host.{.hostname}`
or
`fleet.results.{.uuid}`

With function added into the templating it would be easy to do fanout, sites, label, and more.  This would take some more investigation.

server/service/service.go
```diff
@@ -52,49 +45,28 @@ func NewService(ds kolide.Datastore, resultStore kolide.QueryResultStore,
...
-	osqueryResultLogWriter io.Writer
-
+	resultsQueues   []kolide.QueueService
+	statusQueues    []kolide.QueueService
 	mailService     kolide.MailService
 	ssoSessionStore sso.SessionStore
 	metaDataClient  *http.Client
 }
 
+func (s service) HealthCheckers() (map[string]interface{}) {
```
jrossi: To be honest this needs to be deleted and not done here.  I will fix that by moving this back to when the connections to services are created.

server/kolide/queue.go
```diff
@@ -0,0 +1,10 @@
+package kolide
+
+import (
+	"context"
+	"encoding/json"
+)
+
+type QueueService interface {
```
jrossi: Thought this was a simple interface for results and status and seams correct and would work for just about anything.

server/kolide/health.go
```diff
@@ -0,0 +1,6 @@
+package kolide
+
+
+type HealthService interface {
```
jrossi: needs to be killed

server/queue/nats.go
zwass: This is a really interesting idea that I'd love to hear more about. Can you give some practical examples of what the configuration would look like?

jrossi: Some of the more simple idea: 
```
osquery: 
    results: 
        destination: nats
        nats:
              topic: fleet.platform.{.Platform}.host.{.HostName}
```

With a Nats wildcard subscription you could do things like the following: 

* subscribe to `fleet.platform.darwin.host.>` for all darwin systems 
* subscribe to `fleet.platform.*.host.server-a` for just server-a results 

With the idea of added fucntions something like this would be possiable: 

```
 topic: fleet.company.{ . | label_startswith "firm-" | item 0 }
```
This would find the labels for the host sending the results and then add to the topic.  Something like `fleet.company.firm-jeremyrossi-com`.  *WARNING* I have not look into the details of this and might not make sense with out me knowing more about how labels work. 

Does this show how a templates could be useful?  They would also be useful for kafka, nsq, and other queues available.

jrossi: Some Docs for clients: 
* https://github.com/nats-io/go-nats#wildcard-subscriptions
* https://github.com/nats-io/go-nats#queue-groups <- Not implmeted in my POC Yet

jrossi:
Closing this POC as it was a shared WIP, but I have updated #1875 with my thought on the correct way of adding this features set into fleet.

PR #1916, opened 2018-09-05
Added KeepAlive and ThrottleInterval

golbiga:
Added KeepAlive so the service does not stop on reboot. Also added ThrottleInterval to match com.facebook.osqueryd.plist


groob:
Thanks for the PR. This looks good to me.

Issue #1930, opened 2018-09-27
Support loading Root CA's into net/http TLSClientConfig


adamenger:
Hi! 

We are running an internal PKI using Vault and we've set up Fleet as one of the initial services to test it with. Unfortunately, when I'm trying to log into fleet via the newly minted `fleetctl`, I'm unable to do so because our fleet instance is fronted by a certificate generated by Vault. Even though I have the root certificate added to my workstation, `fleetctl` issues an error when attempting to login:

```
> fleetctl login
Log in using the standard Fleet credentials.
Email: aenger@domain.com
Password:
error logging in: POST /api/v1/kolide/login: Post https://fleet.domain.com/api/v1/kolide/login: x509: certificate signed by unknown authority
```

Would you support loading in a root ca that we can append to the system pool?

PR #1931, opened 2018-09-27
Added rootca support

adamenger:
This attempts to address #1930. I have a root certificate at `/etc/vault/ssl/dev.pem` and I set the config via `fleetctl config set --rootca /etc/vault/ssl/dev.pem`. Once I do this, I'm able to login to my fleet instance which is fronted by a certificated generated through Vault:

```
➜  fleetctl config set --rootca /etc/vault/ssl/dev.pem
➜  ./build/fleetctl login
Log in using the standard Fleet credentials.
Email: aenger@etsy.com
Password:
[+] Fleet login successful and context configured!
```


cmd/fleetctl/api.go
```diff
@@ -32,6 +35,10 @@ func unauthenticatedClientFromCLI(c *cli.Context) (*service.Client, error) {
 		return nil, errors.Wrap(err, "error creating Fleet API client handler")
 	}
 
+	if cc.RootCA != "" {
+		fleet = setupRootCA(fleet, cc.RootCA)
```
zwass: Instead of a separate function, let's do this root CA setup in service.NewClient.

cmd/fleetctl/api.go
```diff
@@ -58,3 +65,26 @@ func clientFromCLI(c *cli.Context) (*service.Client, error) {
...
+	// get system pool, use empty on error
+	rootCAs, _ := x509.SystemCertPool()
+	if rootCAs == nil {
+		rootCAs = x509.NewCertPool()
+	}
+
+	// read in the root cert file specified in the context
+	certs, err := ioutil.ReadFile(rootca)
+	if err != nil {
+		fmt.Println(fmt.Sprintf("Failed to append %v to RootCAs: %v", rootca, err))
```
zwass: fmt.Printf

cmd/fleetctl/config.go
```diff
@@ -215,6 +221,13 @@ func configSetCommand() cli.Command {
 				Destination: &flTLSSkipVerify,
 				Usage:       "Skip TLS certificate validation",
 			},
+			cli.StringFlag{
+				Name:        "rootca",
+				EnvVar:      "ROOTCA",
+				Value:       "",
+				Destination: &flRootCA,
+				Usage:       "Root CA",
```
zwass: Let's be more clear about the purpose of this flag.

server/service/client.go
```diff
@@ -16,7 +16,7 @@ type Client struct {
 	addr               string
 	baseURL            *url.URL
 	token              string
-	http               *http.Client
+	Http               *http.Client
```
zwass: Moving setup to `service.NewClient` will allow us to leave this private.

cmd/fleetctl/api.go
```diff
@@ -58,3 +65,26 @@ func clientFromCLI(c *cli.Context) (*service.Client, error) {
 
 	return fleet, nil
 }
+
+func setupRootCA(client *service.Client, rootca string) *service.Client {
+	// get system pool, use empty on error
+	rootCAs, _ := x509.SystemCertPool()
```
zwass: Can we reason a bit more about whether this functionality should _append_ a cert to the system pool, or use _only_ the certs provided in the pem? I think it may be better to use only the provided cert (and error if this is missing) because it effectively allows pinning of the root.

adamenger:
Hi @zwass, i've addressed your comments and signed the CLA! Please let me know how this looks. One thing I'm unsure of is if my approach of adding another argument to `service.NewClient` is an acceptable approach. Will the rest of your invocations of that client in the codebase need to also pass in a rootCA now too? Also, would you like me to trap the error on line 35 in `server/service/client.go` ?

server/service/client.go
```diff
@@ -30,9 +32,29 @@ func NewClient(addr string, insecureSkipVerify bool) (*Client, error) {
...
 
+	rootCAPool, _ := x509.SystemCertPool()
+	if rootCA != "" {
+		// set up empty cert pool
+		rootCAPool = x509.NewCertPool()
+
+		// read in the root cert file specified in the context
+		certs, err := ioutil.ReadFile(rootCA)
+		if err != nil {
+			fmt.Printf("Failed to append %v to RootCAs: %v\n", rootCA, err)
```
zwass: We need to propagate the error here, as otherwise we will continue with an empty cert chain. Please use `errors.Wrap` (many examples through this file).

server/service/client.go
```diff
@@ -30,9 +32,29 @@ func NewClient(addr string, insecureSkipVerify bool) (*Client, error) {
...
+		}
+
+		// add certs to new cert pool
+		if ok := rootCAPool.AppendCertsFromPEM(certs); !ok {
+			fmt.Printf("No certs appended, using system certs only\n")
```
zwass: same here

adamenger:
Hi @zwass

 I've added the `errors.Wrap()` and I also added some additional handling after line `35` like I mentioned previously just to be extra safe. Let me know what you think. 

Thanks for your guidance,

adamenger:
Ah I see CI failures, fixing...

adamenger:
Thank you for your guidance!

PR #1935, opened 2018-10-08
Note regarding kolide.crt

lgalvez338:
Having a mismatched certificate name when creating the kolide-enroll package causes TLS handshake errors. Additionally, the end point running osqueryd reports a transport error "x509: certificate signed by unknown authority" which is sort of a red herring.

### DIFF
--- a/docs/infrastructure/adding-hosts-to-fleet.md
+++ b/docs/infrastructure/adding-hosts-to-fleet.md
@@ -137,6 +137,8 @@ Next, you'll have to edit the `config.mk` file. You'll find all of the necessary
 
 Note that osqueryd requires a full certificate chain, even for certificates which might be trusted by your keychain. The "Fetch Kolide Certificate" button in the Add New Host screen will attempt to fetch the full chain for you.
 
+If your server certificate is not named kolide.crt, change the certificate name variable in --tls_server_certs to match your server certificate.
+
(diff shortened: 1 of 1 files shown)


zwass:
Can you help me understand more about the motivation of this commit? In the context of that section of documentation, the certificate is always named `kolide.crt` (because the Makefile writes it as that filename).

lgalvez338:
Line 108 references making a change based on a different search path. I think it might be a good idea to make a similar reference for in this section, for those not be familiar with makefile, who might overlook changing the certificate name and path.

docs/infrastructure/adding-hosts-to-fleet.md
```diff
@@ -137,6 +137,8 @@ Next, you'll have to edit the `config.mk` file. You'll find all of the necessary
 
 Note that osqueryd requires a full certificate chain, even for certificates which might be trusted by your keychain. The "Fetch Kolide Certificate" button in the Add New Host screen will attempt to fetch the full chain for you.
 
+If your server certificate is not named kolide.crt, change the certificate name variable in --tls_server_certs to match your server certificate.
```
marpaia, line 140: It's not clear to me what program I would pass the `--tls_server_certs` flag. I happen to know that's an `osqueryd` flag, but it is not clear from this excerpt what certificate we're talking about and when I should use this flag.

lgalvez338: Sorry for the confusion. I mean to say edit the config.mk file line "--tls_server_certs=/etc/osquery/kolide.crt" to match your TLS certificate name instead of the default kolide.crt.

zwass:
Closing due to inactivity.

PR #1947, opened 2018-10-31
Documentation: Add mysql conn limits 

robusto:
The 'max_open_conns' and 'max_idle_conns' settings seem to be missing. I was referencing the config struct in 'server/config/config.go'


groob:
Good one!

PR #1957, opened 2018-11-19
Fix errors when trying to retrieve specs with spaces in name

### LINKED ISSUE #1948: fleetctl: unable to find entities with spaces in the name
```
fleet - version 2.0.0
  branch: 	HEAD
  revision: 	3a07a4c07e5c216c5680a0b0dd5300f6019bf7d6
  build date: 	2018-10-25T20:00:12Z
  build user: 	marpaia
  go version: 	go1.11
```
alpine

### What did you do?

1) Create / apply a query with spaces:
```
/ # cat example.yml 
apiVersion: v1
kind: query
spec:
  name: query name with spaces
  description: some description
  query: select * from docker_containers as dc join docker_container_stats as dcs
    on dc.id=dcs.id;
/ # fleetctl apply -f example.yml 
[+] applied 1 queries
```
At this point, the query has been created and can be seen in the UI.

2) Try to delete the query:
```
/ # fleetctl delete -f example.yml
[+] deleting query "query name with spaces"
[!] query "query name with spaces" doesn't exist
```
I expected the query `query name with spaces` to be deleted.
### What did you see instead?
The query `query name with spaces` was not found and continues to exist. I can manually delete through the UI but not through the cli.

Not demonstrated, but I have tested this with labels and get the same problem. I haven't tested with packs.

zwass:
We need to properly escape and unescape the name parameter.

Fixes #1948

### DIFF
--- a/server/service/transport.go
+++ b/server/service/transport.go
@@ -3,13 +3,14 @@ package service
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"io"
 	"net/http"
+	"net/url"
 	"strconv"
 
 	"github.com/gorilla/mux"
 	"github.com/kolide/fleet/server/kolide"
+	"github.com/pkg/errors"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


server/service/transport.go
```diff
@@ -74,7 +75,11 @@ func nameFromRequest(r *http.Request, varName string) (string, error) {
 	if !ok {
 		return "", errBadRoute
 	}
-	return name, nil
+	unescaped, err := url.PathUnescape(name)
```
zwass, line 78: I verified this function is only called from the routes we intend to change, so this should not have unintended effects on other routes.

PR #1959, opened 2018-11-20
Delete soft-deleted entities

### LINKED ISSUE #1956: Fleetctl Returning Inaccurate Pack YAML
```
{
  "version": "2.0.0",
  "branch": "HEAD",
  "revision": "3a07a4c07e5c216c5680a0b0dd5300f6019bf7d6",
  "go_version": "go1.11",
  "build_date": "2018-10-25T20:00:12Z",
  "build_user": "marpaia"
}
```
Probably worth noting that this console has been updated quite a bit over time, so there may be some legacy DB artifacts from older builds hanging around.

Linux/Docker to host Fleet

### What did you do?
`fleetctl get pack macpack`

A 1:1 representation of what I see in the Fleet UI in the YAML

### What did you see instead?
Queries that are no longer present in the pack (and some which have been deleted from the console entirely) are being included in the pack YAML.

I've been able to semi-work around this problem by re-creating deleted queries, adding them, and then immediately removing them from the pack. It seems whatever datasource Fleetctl pulls from gets updated when I do that. 

Here's the most straight forward example I can provide:
Our console shows 53 queries present for the "macpack" query pack:
image

However, using fleetctl returns 61 queries:
```
$ fleetctl get pack macpack | grep name: | grep -v 'name: macpack' | wc -l
61
```

zwass:
Avoids potential bugs in which soft-deleted entities are returned from database
queries (soft-deletion is now deprecated), but some records may still exist.

Fixes #1956

### DIFF
--- a/server/datastore/mysql/migrations/data/20181119180000_DeleteSoftDeletedEntities.go
+++ b/server/datastore/mysql/migrations/data/20181119180000_DeleteSoftDeletedEntities.go
@@ -1,0 +1,54 @@
+package data
+
+import (
+	"database/sql"
+
+	"github.com/pkg/errors"
+)
+
+func init() {
+	MigrationClient.AddMigration(Up20181119180000, Down20181119180000)
+}
+
+func Up20181119180000(tx *sql.Tx) error {
+	sql := `DELETE FROM scheduled_queries WHERE deleted = 1`
+	if _, err := tx.Exec(sql); err != nil {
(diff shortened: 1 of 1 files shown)


zwass:
I'm looking at the schema and see that `queries.author_id` is set to null on deletion. Scheduled queries cascade deletion with associated query and associated pack. Can you think of other entities that should have cascading deletion?

I'd probably want to handle those separately from this PR, but glad to discuss here.

PR #1965, opened 2018-12-08
Fix slow return of results in fleetctl query with high host count

### LINKED ISSUE #1895: fleet returning results one by one instead of fetching all available results at the same time
Following discussion on slack with @zwass 

After executing a query (UI and fleetctl) I'm seeing results returned one by one instead of all at the same time.
While examining redis instance I can see that all my osquery clients already responded to the query, but fleet is fetching them gradually instead of fetching all available results from redis.
Also, this is clearly visible when looking at current results number, I'm getting 1-3 results per second.
I would expect to get all available results from redis.
Since picture is worth a thousand words I've attached screenshot showing redis memory usage after running the query via fleetctl.
Clearly visible spike represents a moment when results are returned from osquery agents, next we can observe that fleet is slowly fetching results - one by one.
Unfortunately with the number of hosts we're planning to use fleet with - this will not scale.

redis

When looking at MySQL's `distributed_query_executions` table I can confirm the same - results from osquery daemons are available almost instantly, while fleet UI and fleetctl is returning them one by one.

This is regardless of using websockets or not.

`2.0.0-rc3`

linux

### What did you do?
ran a query

All available results returned at the same time

### What did you see instead?
Results are returned one by one

zwass:
Prior to this change, the live query status was calculated after each retrieved
result. This was fine at a low host count, but when more hosts are added, doing
this calculation becomes slower.

Now, the status is only calculated every 5 seconds, while results can be
streamed continuously.

Fixes #1895

### DIFF
--- a/server/service/service_campaigns.go
+++ b/server/service/service_campaigns.go
@@ -150,22 +150,18 @@ func (svc service) StreamCampaignResults(ctx context.Context, conn *websocket.Co
 		}
 	}
 
-	// Loop, pushing updates to results and expected totals
-	for {
-		// Update the expected hosts total (Should happen before
-		// any results are written, to avoid the frontend showing "x of
-		// 0 Hosts Returning y Records")
+	updateStatus := func() error {
 		hostIDs, labelIDs, err := svc.ds.DistributedQueryCampaignTargetIDs(campaign.ID)
 		if err != nil {
 			if err = conn.WriteJSONError("error retrieving campaign targets"); err != nil {
-				return
+				return errors.New("retrieve campaign targets")
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


server/service/service_campaigns.go
```diff
@@ -190,9 +186,26 @@ func (svc service) StreamCampaignResults(ctx context.Context, conn *websocket.Co
...
+		return nil
+	}
+
+	if err := updateStatus(); err != nil {
+		svc.logger.Log("msg", "error updating status", "err", err)
+		return
+	}
+
+	// Push status updates every 5 seconds at most
+	tick := time.NewTicker(5 * time.Second)
```
groob: nit. use `ticker` instead of `tick`

PR #1969, opened 2018-12-20
Non-Quoted Interval Conversion

jordan-rash:
Found that when running `fleetctl convert -f ./mypack.yml`, observed that any of the non-quoted integer values (read as strings) were being converted to 0 in the final output.  Ran a simple test to determine when a number is Unmarshalled as an interface (https://github.com/kolide/fleet/blob/06832697d0e6ed6b2ca0220ef5434791db7b0a27/server/kolide/osquery.go#L52) it is of type `float64`.  Simply added a `float64` to the case statement and now non-quoted integers are being rendered correctly.


zwass:
Nice, thanks Jordan!

PR #1971, opened 2018-12-21
Add username support to fleetctl setup

### LINKED ISSUE #1970: Feature Request: Allow Username to be set with fleetctl
2.0.1

MacOS

### What did you do?
When configuring a user via `fleetctl setup` for the first time, there doesn't seem to be an option to configure a username for the user.

It would be dope if there was an option in `fleetctl setup` to allow a user to provide a username when configuring the initial user. It would even be fine if the first user was named "admin" by default.

zwass:
Username continues to default to email if not specified in the options.

Closes #1970

### DIFF
--- a/cmd/fleetctl/setup.go
+++ b/cmd/fleetctl/setup.go
@@ -13,13 +13,14 @@ import (
 func setupCommand() cli.Command {
 	var (
 		flEmail    string
+		flUsername string
 		flPassword string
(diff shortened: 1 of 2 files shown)


zwass:
I manually tested that this allows setting a separate username in `fleetctl setup`, and that the normal web UI setup works as before.

PR #1972, opened 2018-12-27
Allow selection of query targets by UUID in URL params

### LINKED ISSUE #1934: Pre-select host by IP for ad-hoc queries
I am looking for a way to pre-select a host by IP for an ad-hoc query. I see that I can pre-select using the host ID `https://FleetInstall/queries/new?host_ids=20` But I am not seeing a way to use IP. 

This is to facilitate easier pivoting to Fleet from another tool.

zwass:
Closes #1934

### DIFF
--- a/frontend/pages/queries/QueryPage/QueryPage.jsx
+++ b/frontend/pages/queries/QueryPage/QueryPage.jsx
@@ -86,7 +86,7 @@ export class QueryPage extends Component {
   componentWillMount () {
     const { dispatch, hostIDs, hostUUIDs, selectedHosts, selectedTargets } = this.props;
 
-    if (hostIDs || hostUUIDs) {
+    if (((hostIDs && hostIDs.length) || (hostUUIDs && hostUUIDs.length)) > 0) {
       dispatch(hostActions.loadAll());
     }
 
@@ -573,7 +573,7 @@ const mapStateToProps = (state, ownProps) => {
   const title = queryID ? 'Edit Query' : 'New Query';
   let selectedHosts = [];
 
-  if (!queryID && (hostIDs || hostUUIDs)) {
+  if (!queryID && ((hostIDs && hostIDs.length) || (hostUUIDs && hostUUIDs.length)) > 0) {
     const hostIDsArr = Array.isArray(hostIDs) ? hostIDs : [hostIDs];
     const hostUUIDsArr = Array.isArray(hostUUIDs) ? hostUUIDs : [hostUUIDs];
     const { entities: hosts } = stateEntities.get('hosts');
--- a/frontend/pages/queries/QueryPage/QueryPage.jsx
+++ b/frontend/pages/queries/QueryPage/QueryPage.jsx
@@ -2,7 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import classnames from 'classnames';
 import { connect } from 'react-redux';
 import FileSaver from 'file-saver';
-import { clone, filter, includes, isArray, isEqual, merge } from 'lodash';
+import { clone, concat, filter, includes, isArray, isEqual, merge, uniq } from 'lodash';
 import moment from 'moment';
 import { push } from 'react-router-redux';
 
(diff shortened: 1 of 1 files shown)


frontend/pages/queries/QueryPage/QueryPage.jsx
```diff
@@ -83,9 +84,9 @@ export class QueryPage extends Component {
   }
 
   componentWillMount () {
-    const { dispatch, hostIDs, selectedHosts, selectedTargets } = this.props;
+    const { dispatch, hostIDs, hostUUIDs, selectedHosts, selectedTargets } = this.props;
 
-    if (hostIDs) {
+    if (hostIDs || hostUUIDs) {
```
kyleknighted: Should we verify if it's an array that has data?

frontend/pages/queries/QueryPage/QueryPage.jsx
```diff
@@ -567,24 +568,25 @@ const mapStateToProps = (state, ownProps) => {
...
-      ? h => includes(hostIDs, String(h.id))
-      : { id: Number(hostIDs) };
-
+  if (!queryID && (hostIDs || hostUUIDs)) {
+    const hostIDsArr = Array.isArray(hostIDs) ? hostIDs : [hostIDs];
+    const hostUUIDsArr = Array.isArray(hostUUIDs) ? hostUUIDs : [hostUUIDs];
     const { entities: hosts } = stateEntities.get('hosts');
+    // hostIDs are URL params so they are strings and comparison with ints may
+    // need conversion.
+    const hostFilter = h => includes(hostIDsArr, String(h.id)) || includes(hostUUIDsArr, String(h.uuid));
```
kyleknighted, line 582: If you're replacing lodash functionality with native for `isArray`, we could do the same for [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

frontend/pages/queries/QueryPage/QueryPage.jsx
zwass: Unfortunately (IMO) the JS API gives us a string if a single parameter was provided, or an array of strings if multiple parameters were provided. Do you think there's a good way to check for that?

frontend/pages/queries/QueryPage/QueryPage.jsx
zwass: What do you think the tradeoffs are? Should I always prefer lodash? I see that `isArray` is compatible with every browser on MDN, while `includes` is compatible with all besides IE.

frontend/pages/queries/QueryPage/QueryPage.jsx
kyleknighted: I think this should work. `.length` is a proto on both strings and arrays - this conditional should check for any length on either variable. Otherwise, something like an erroneous empty array would trigger a truthy result in the current conditional.

```suggestion
    if ((hostIDs.length || hostUUIDs.length) > 0) {
```

frontend/pages/queries/QueryPage/QueryPage.jsx
kyleknighted: It's being compiled with Babel, so we can make sure it supports the required browsers - https://babeljs.io/docs/en/babel-preset-env#targets

zwass: Nice, we have the browsers specified in https://github.com/kolide/fleet/blob/master/package.json#L118-L124, so I think we're good here?

frontend/pages/queries/QueryPage/QueryPage.jsx
zwass: I used a modified version of this because these variables are sometimes undefined (when no argument is specified at all) and would throw exceptions.

frontend/pages/queries/QueryPage/QueryPage.jsx
kyleknighted: It's really up to you. Maybe it's just a followup PR where we reassess our uses of lodash?

zwass: To finalize this, Kyle and I connected over chat and we believe that the Babel configuration will handle appropriate browser support.

PR #1977, opened 2019-01-03
Add 'get hosts' command to fleetctl

### LINKED ISSUE #1962: Add fleetctl get hosts command
Hello ,

This is a feature request for fleetctl! 

Currently I am using fleetctl with the following versioning information:

```
$ fleetctl -version
fleetctl - version 2.0.0
  branch: 	master
  revision: 	3a07a4c07e5c216c5680a0b0dd5300f6019bf7d6
  build date: 	2018-10-16T22:32:34Z
  build user: 	zwass
  go version: 	go1.11
```

on a redhat variant machine.

fleetctl currently does not provide a good way of getting a listing of OFFLINE hosts, ONLINE hosts, MIA hosts. We also can't get a count of these hosts. 

This request is similiar to mimicing the Kolide UI that when you login and clicking the OFFLINE hosts you would see all OFFLINE hosts. You can see the number of OFFLINE hosts as well prior clicking the link. 

I have not been able to do the same thing with fleetctl.. If I knew for example a type of query the UI does for that, maybe I could do it that way, but it does not seem to be the case.

As an example, I want to do something similar to this where I can make a query to get a listing :

`fleetctl query --query 'SELECT * FROM  offline_hosts"' --labels='All Hosts' `
or mia_hosts, or online_hosts. Maybe number of hosts as well..  

If it helps please see the discussion I had with @blaedj in the Kolide Slack channel about this:

https://osquery.slack.com/archives/C1XCLA5DZ/p1543277266660300

Thank you!

blaedj:
This commit:
 - adds a new subcommand for fetching hosts to `fleetctl get` command.

Why?
 - this allows for listing of all hosts via the fleetctl interface.
 There may be additional attributes of the host that we'd like to
 display, but this should be a good start.

Closes #1962

### DIFF
--- a/cmd/fleetctl/get.go
+++ b/cmd/fleetctl/get.go
@@ -326,7 +326,7 @@ func getHostsCommand() cli.Command {
 
 			for _, host := range hosts {
 				data = append(data, []string{
-					host.Host.UUID, // not sure if this is the most appropriate value to display as the 'id'
+					host.Host.UUID,
 					host.DisplayText,
 					host.Host.Platform,
 					host.Status,
@@ -334,7 +334,7 @@ func getHostsCommand() cli.Command {
 			}
 
 			table := defaultTable()
-			table.SetHeader([]string{"id", "hostname", "platform", "status"})
+			table.SetHeader([]string{"uuid", "hostname", "platform", "status"})
 			table.AppendBulk(data)
 			table.Render()
 
--- a/cmd/fleetctl/get.go
+++ b/cmd/fleetctl/get.go
@@ -296,3 +296,49 @@ func getEnrollSecretCommand() cli.Command {
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


cmd/fleetctl/get.go
```diff
@@ -296,3 +296,49 @@ func getEnrollSecretCommand() cli.Command {
...
+				data = append(data, []string{
+					host.Host.UUID, // not sure if this is the most appropriate value to display as the 'id'
+					host.DisplayText,
+					host.Host.Platform,
+					host.Status,
+				})
+			}
+
+			table := defaultTable()
+			table.SetHeader([]string{"id", "hostname", "platform", "status"})
```
zwass: If we are using UUID, I think the header should be `uuid`.

PR #1984, opened 2019-01-10
Fix typo in query yaml example

crimsonknave:


crimsonknave:
It appears that `fleetctl` requires you to specify a `name` for each query in a query pack (like the two different version queries have). I'm happy to update the example in the file to reflect this, but I wasn't sure if that was a bug (and I should open up an issue) or intended behavior.

crimsonknave:
So, I think it's pretty reasonable to have it default to the query name. 

I'd also add in that if `fleetctl` doesn't default the name it should have an error that says that the query name is required. Currently, the server just returns a database error about duplicate names (below) which took me a bit of digging to figure out what was going on.

`applying packs: apply packs received status 500 Unknown Error: applying pack 'osquery/monitoring': adding query  referencing osquery/monitoring/events: Error 1062: Duplicate entry '8' for key 'unique_names_in_packs'`

zwass:
I created an issue to fix the behavior as discussed: https://github.com/kolide/fleet/issues/1990

Going to merge this doc fix now. Thank you.

Issue #1663, opened 2017-12-08
Sending mail: Client Auth Error


lucasjhall:
### General Issue 
When setting up smtp relay for Office 365 the following error was received:
```sending mail: client auth error: 504 5.7.4 Unrecognized authentication type [CY4PR0601CA0001.namprd06.prod.outlook.com]```

![client-auth-err](https://user-images.githubusercontent.com/8007807/33785927-021a2a1e-dc1c-11e7-8a5a-721c1dfb5ccb.png)

I did a little searching and there seemed to be a similar issue in another project [gophish/244](https://github.com/gophish/gophish/issues/244)

#### Settings
SMTP Server: `smtp.office365.com`
Port: `587`
Username and Password: `Y`
Auth Mehod: `Plain`
#### Advanced Settings 
Verify SSL Certs? `ON`
Enable STARTTLS? `ON`

### Fleet Version
``` 
branch: 	master
  revision: 	45165aa29aae52346ed1c458f19b89947d315c61
  build date: 	2017-12-04T22:50:27Z
  build user: 	marpaia
  go version: 	go1.9.2
```
### OS 
`Ubuntu 16.04`

**Comment by zwass** [comment]
I can repro this using a free outlook.com address. Would be happy to review a PR if someone would like to resolve it. Modifications will probably go in https://github.com/kolide/fleet/blob/master/server/mail/mail.go.

**Comment by Jackbennett** [comment]
Onprem exchange I'm getting the same issue. Similar config. Fleet version from docker, same fleet version.

**Comment by lucasjhall** [comment]
I just verified this still exists in 2.0.0-rc3 8b7edf7

**Comment by zwass** [comment]
We'd happily accept a PR for this if someone submits one. We don't have any O365 or Exchange to test on.

If it helps motivate you, here's an email auth PR (https://github.com/kolide/fleet/pull/1805) that went from submitted to merged in a single day.

**Comment by moudis** [comment]
I know this is an older issue, but during setup of a Kolide Fleet instance I came across this. Office 365 Exchange only allows LOGIN and XOAUTH2 authentication methods (instead of PLAIN or CRAM-MD5). From a session with the server:

```
250 SMTPUTF8
EHLO smtp.office365.com
250-CO2PR04CA0123.outlook.office365.com Hello [2602:61:7137:ea00:99ff:927f:36a1:57cf]
250-SIZE 157286400
250-PIPELINING
250-DSN
250-ENHANCEDSTATUSCODES
[...]

PR #1988, opened 2019-01-11
Add Support for SMTP LOGIN Authentication Method

moudis:
#1663 

This PR adds support for the SMTP LOGIN authentication method. Office 365 Exchange removed support for PLAIN authentication some time ago, and only supports LOGIN and an OAuth2 authentication method. This patch has been tested with a licensed O365 account. This method should also be usable with any other email server that advertises LOGIN in its 250-AUTH response.

Note: If using this with O365, the account used must not have MFA enabled.


zwass:
Actually the message did eventually come through from my test.

PR #1989, opened 2019-01-12
Add --timeout flag to fleetctl query

### LINKED ISSUE #1987: Add --timeout flag to fleetctl query
Users are requesting a `--timeout` flag for the `fleetctl query` command. This flag would allow the query to run for some time before terminating the query process.

zwass:
Closes #1987


cmd/fleetctl/query.go
```diff
@@ -148,6 +165,10 @@ func queryCommand() cli.Command {
 						}
 						return nil
 					}
+
+				// Check for timeout expiring
+				case <-timeoutChan:
+					return nil
```
marpaia, line 175: Perhaps emit a log here?

PR #1992, opened 2019-01-16
Add FAQ for Fleet deployments

zwass:


docs/infrastructure/faq.md
```diff
@@ -0,0 +1,81 @@
...
+- Ensure that the CNAME on the certificate matches the address at which the server is being accessed. If I try connect osquery via `https://localhost:443`, but my certificate is for `https://fleet.example.com`, the verification will fail.
+- Is Fleet behind a load-balancer? Ensure that if the load-balancer is terminating TLS that this is the certificate provided to osquery.
+- Does the certificate verify with `curl`? Try `curl -v -X POST https://kolideserver:port/api/v1/osquery/enroll`.
+
+## What do I do about "too many open files" errors?
+
+This error usually indicates that the Fleet server has run out of file descriptors. Fix this by increasing the `ulimit` on the Fleet process. See the `LimitNOFILE` setting in the [example systemd unit file](./systemd.md) for an example of how to do this with systemd.
+
+## The UI constantly logs me out with 500 errors, or logs indicate "too many connections". How do I fix this?
```
groob: Actually, because we enforce a max_connections, this error is not likely to happen anymore. The more common case is that fleet is running slow, even though the user has upgraded to a more performant database tier, but the fleet flags are still at their default values (of 50 conns).

PR #1998, opened 2019-01-23
Hide labels for platforms with no hosts.

zwass:
Rather than displaying 0 hosts for each of the platforms, remove the platform
listing. The platforms will appear once they have a host.


zwass:
Before:
screen shot 2019-01-23 at 1 49 50 pm


After:
screen shot 2019-01-23 at 1 49 06 pm

PR #2009, opened 2019-03-06
`config_tls_refresh` => `config_refresh` in docs

crimsonknave:
Documentation has a typo. The field is `config_refresh`.


zwass:
The flags are aliased, so both are correct. Typically docs these days refer to `config_refresh`, so I agree that this is a good idea. Can you please change it in the other locations where it appears in the docs? Thank you!

crimsonknave:
I seem to recall that things weren't changing until I changed the name to the field, so it's possible that they aren't aliased in older versions of osquery. My hosts are all running `3.2.4` currently.

I can certainly update it in the rest of the docs.

zwass:
In osquery versions before 2.4.6, `config_refresh` did not exist. Since then `config_tls_refresh` has been aliased to  `config_refresh`. Typically folks are using `config_refresh` these days. There are a couple more spots I found where we can change this. I'll push a commit for that and then merge this.

PR #2022, opened 2019-04-04
Add Firehose logging capabilities for result and status logs

zwass:
This PR allows osquery status/result logs to be independently configured to forward from Fleet to AWS Firehose. Status and result streams can be independently configured to go to the filesystem or to different streams in Firehose. The pattern used leaves room for additional log sinks to be introduced later.


terracatta:
@zwass Excited to try this out. If you could include some basic instructions I can try it on a fresh Firehose setup on AWS.

I noticed you added some new config options to the config struct relating to "streams" which I assume is a Firehose term.

```go
	StatusLogStream     string        `yaml:"status_log_stream"`
	ResultLogStream     string        `yaml:"result_log_stream"`
```

Do you think perhaps options that relate to a particular plugin should be named something like `firehose_status_log_stream` and `firehose_result_log_stream` to give an end-user a sense of the dependency relationship between a set of options and the plugin they might depend on?

zwass:
This is a great comment regarding the flags. My thinking was that it seems likely someone will add a Kafka logger, etc. at some point and then that flag can apply to the stream name to use in those other systems as well. Does that seem more confusing than it is worth?

terracatta:
> This is a great comment regarding the flags. My thinking was that it seems likely someone will add a Kafka logger, etc. at some point and then that flag can apply to the stream name to use in those other systems as well. Does that seem more confusing than it is worth?

I think it could be confusing because if we were talking about kafka I would imagine the correct terminology would be "topics" instead of "streams", right? This sort of sets a precedent for future plugins that they should re-use config options that are "close enough" sounding to what they actually need.

I think we should namespace them. Even if it means a little more work for a user to change up their config when switching from one plugin to another.

terracatta:
@zwass This PR is setting us up for one logger plugin active at a time. Do you think users will demand the ability to run multiple logger plugins simultaneously? My fear is the community creates a lot of great logging plugins, but people just use the filesystem one anyway because they want the ability to divert logs to multiple log aggregation systems and that would be the only way to do it.

I am curious though if you think that use-case is actually not practical and should be ignored.

server/logging/firehose.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/firehose"
+	"github.com/aws/aws-sdk-go/service/firehose/firehoseiface"
+	"github.com/pkg/errors"
+)
+
+const (
+	firehoseMaxRetries = 3
```
terracatta: Seems weird to me that you have this exponential back-off strategy but only try 3 times. Any reason why this is so low?

server/logging/firehose.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if (*(*out.DeliveryStreamDescription).DeliveryStreamStatus) != firehose.DeliveryStreamStatusActive {
+		return errors.Errorf("delivery stream %s not active", f.stream)
+	}
+
+	return nil
+}
+
+func (f *firehoseLogWriter) Write(logs []json.RawMessage) error {
```
directionless, line 74: I found this write loop fairly hard to understand. Maybe it's a common idiom, I'm not sure. I'd add comments to clarify.

If I read it right, this is:
* trying to write at MaxRecord size
* so it looping over logs
* _If_ it `>= firehoseMaxRecordsInBatch`, then it attempts a write
* if not, it keeps looping appending logs
* writes things when the loop is over.

I imagine you're doing this, instead of iterating by N, since you also need to track firehoseMaxSizeOfBatch

server/logging/firehose.go
directionless: agreed/

server/logging/filesystem.go
```diff
@@ -0,0 +1,138 @@
...
+// NewFilesystemLogWriter creates a log file for osquery status/result logs the
+// logFile can be rotated by sending a `SIGHUP` signal to kolide if
+// enableRotation is true
+func NewFilesystemLogWriter(path string, appLogger log.Logger, enableRotation bool) (*filesystemLogWriter, error) {
+	if enableRotation {
+		// Use lumberjack logger that supports rotation
+		osquerydLogger := &lumberjack.Logger{
+			Filename:   path,
+			MaxSize:    500, // megabytes
+			MaxBackups: 3,
```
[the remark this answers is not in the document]
zwass: This is just moved code from the existing implementation. Seems totally reasonable to me to make this configurable but I'd say it's out of scope for this PR.

server/logging/firehose.go
```diff
@@ -0,0 +1,141 @@
...
+	var records []*firehose.Record
+	totalBytes := 0
+	for _, log := range logs {
+		if len(log) > firehoseMaxSizeOfRecord {
+			// TODO zwass: better logging
```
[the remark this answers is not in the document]
zwass: Yeah... this is a hard limit for Firehose and it mimics the behavior of osquery in the case of logs that are too big. At the least I'd like to make this log to stderr some information about the log that is dropped. Maybe it could write oversize logs to a local file?

server/logging/firehose.go
zwass: You seem to have it right. I will add some comments to try to clarify. I'm not sure there's a simpler way to write this loop given the two constraints (count and total size) we need to respect for batching the logs.

zwass:
Regarding config names, I agree with you @terracatta. No need to try to shove multiple concepts into one config.

Regarding logging to multiple sinks, I think you are right that users will almost certainly demand that option. My personal opinion is that any sophisticated splitting of logs should happen in the logging pipeline, but I think this implementation leaves room to add the ability to log all logs to multiple sinks.

The tricky thing about multiple sinks is managing duplication in the face of errors. Right now, if we fail to write to Firehose we can return an error to osqueryd and it will retry later with those same logs. What happens if one log sink succeeds while the other fails? Ultimately I think we would return the error and expect the log consumers to handle duplication (fwiw the [Firehose docs](https://docs.aws.amazon.com/firehose/latest/APIReference/API_PutRecordBatch.html) say "We recommend that you handle any duplicates at the destination.").

All in all, my feeling is that this implementation satisfies a current use case and leaves the options open for the future.

server/logging/firehose.go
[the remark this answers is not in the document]
terracatta: >  At the least I'd like to make this log to stderr some information about the log that is dropped. Maybe it could write oversize logs to a local file?

A stderr message sounds good enough. I think logging them to a local file brings up concerns around rotating that file (which you will have to consider because you are in fact writing massive logs to it).

At the end of the day, the goal is for end-users to easily identify the queries generating massive logs. If I were a user and I had result logs that were oversized, I'd want to see the info about the scheduled query producing the large logs and the actual size of the logs in both bytes and in # of rows. I'd also want to see the host UUID any decorators so I could identify the host easily.

server/logging/filesystem.go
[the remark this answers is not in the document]
directionless: WFM

server/logging/firehose.go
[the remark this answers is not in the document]
directionless: I'm with Jason. If the logs are so big they can't go to firehose, storing them on disk is going to raise all kinds of rotation questions. Logging enough for debugging ought be enough.

directionless:
> My personal opinion is that any sophisticated splitting of logs should happen in the logging pipeline, but I think this implementation leaves room to add the ability to log all logs to multiple sinks.

A counter point -- Even if you mostly want to move everything into a pipeline and handle complexity there, one often has multiple pipelines for redundancy. A common setup I've seen is agents pushing to parallel kafaka pipelines, which are dedupped and sent along.

zwass:
@directionless I think we are in agreement, but I'd like to confirm I understand the setup you describe. Are you suggesting that folks push all the logs to both kafka pipelines, or are you suggesting some sort of load-balancing between the pipelines?

directionless:
I've seen both. But more often is pushing all logs to both pipelines for redundancy. (assume some downstream deduplication)

zwass:
I have addressed all of the review comments and extensively updated the documentation. Please let me know if there is other feedback before merging. Thanks!

server/config/config.go
```diff
@@ -86,19 +88,42 @@ type LoggingConfig struct {
 	DisableBanner bool `yaml:"disable_banner"`
 }
 
+// AwsConfig defines configs related to AWS services
+type AwsConfig struct {
```
terracatta: How do you feel about also putting these under firehoseconfig struct? Once again, for the same reason, perhaps the community introduces a separate orthogonal thing related to AWS that also requires creds (stilly straw-man example: pulling in pack files stored in s3) they may want to use different creds.

What do you think?

server/logging/firehose.go
```diff
@@ -0,0 +1,162 @@
...
+				"log", string(log[:100])+"...",
+			)
+			continue
+		}
+
+		// If adding this log will exceed the limit on number of
+		// records in the batch, or the limit on total size of the
+		// records in the batch, we need to push this batch before
+		// adding any more.
+		if len(records) >= firehoseMaxRecordsInBatch ||
```
terracatta: Should this be `> firehoseMaxRecordsInBatch || totalBytes+len(log) > firehoseMaxSizeOfBatch` instead of `>=` ?

terracatta, line 96: Should these be `>` instead of `>=`?

server/logging/firehose.go
```diff
@@ -0,0 +1,162 @@
...
+func (f *firehoseLogWriter) Write(logs []json.RawMessage) error {
+	var records []*firehose.Record
+	totalBytes := 0
+	for _, log := range logs {
+		// We don't really have a good option for what to do with logs
+		// that are too big for Firehose. This behavior is consistent
+		// with osquery's behavior in the Firehose logger plugin, and
+		// the beginning bytes of the log should help the Fleet admin
+		// diagnose the query generating huge results.
+		if len(log) > firehoseMaxSizeOfRecord {
```
terracatta, line 83: Just a question. Have you tried inserting a log 1 byte less than the max to a real firehose server and seeing if it makes it across?

I know it _should_, I just would hate for us to hit an edge case where the library is adding some encoding to the logs behind the scenes that actually causes them to exceed this limit.

server/logging/firehose.go
```diff
@@ -0,0 +1,162 @@
...
+
+		// Not retryable or retries expired
+		return err
+	}
+
+	// Check errors on individual records
+	if output.FailedPutCount != nil && *output.FailedPutCount > 0 {
+		if try >= firehoseMaxRetries {
+			return errors.Errorf(
+				"failed to put %d records, retries exhausted",
```
terracatta: Kind of a shame we can't use `output.RequestResponses` to give the more info on why N records couldn't be pushed. I didn't see it, but is there any point where we make error status codes available to a user?

If we don't, do you think we should?

terracatta:
@zwass This is looking really good. I really like the documentation changes and the deprecation of pre-existing generically named filesystem logger options.

Just a couple more questions mostly...once we get through those and assuming @directionless doesn't see anything glaring, I'd be happy to merge this.

server/logging/firehose.go
zwass: Line 96 needs to be `>=` otherwise we will end up trying to write one more than the max number of records in a batch. Line 97 could be changed to `>` to ensure we write up to the absolute maximum in a single batch.

server/logging/firehose.go
zwass: It's not really clear the best way to expose the potentially up to 500 errors to the user. Maybe the error from a single record would be representative and we could expose that?

server/logging/firehose.go
zwass: I was able to confirm that I could put 4 records of exactly 1MB each and retrieve 4MB on the other end. Seems that the docs are accurate (https://docs.aws.amazon.com/sdk-for-go/api/service/firehose/#Firehose.PutRecordBatch)

> Each PutRecordBatch request supports up to 500 records. Each record in the request can be as large as 1,000 KB (before 64-bit encoding), up to a limit of 4 MB for the entire request. These limits cannot be changed.

server/config/config.go
zwass: I would suggest that it's typically going to be easier to use a single service account with the appropriate permissions.

But I would like to avoid having to do any further deprecation of configs in the future so I will break these out directly into the `FirehoseConfig`.

server/logging/firehose.go
zwass: I exposed the first error.

zwass:
Thanks y'all for the great reviews on this one!

PR #2025, opened 2019-04-09
Close flate writer before base64 encoding SAML authorization request

### LINKED ISSUE #2024: SAML AuthnRequest fails to be inflated by many tools
Hey there! This is Jamie from Duo Security. We have a mutual customer that is running into some issues using the SAML feature of Fleet.

While attempting to test a SAML authentication where Fleet is configured to use [Duo Access Gateway](https://duo.com/docs/dag-linux) (DAG) as its SAML Identity Provider we get an error on the DAG when it attempts to inflate the XML from the AuthnRequest that Fleet sent over. 

An error also happens when you take the base64 deflated XML and try to decode and inflate it using https://www.samltool.com/decode.php and https://chrome.google.com/webstore/detail/saml-chrome-panel/paijfdbeoenhembfhkhllainmocckace?hl=en

Screen Shot 2019-04-08 at 8 39 50 AM
Screen Shot 2019-04-08 at 8 39 18 AM

Both of these tools have never had an issue decoding and inflating SAML request before.

Upon running the deflate code manually in https://github.com/kolide/fleet/blob/7153e0043c7430f46ffa8d766228def413eff56a/server/sso/authorization_request.go#L116 we found that if we tried to inflate it using a zlib in python we got `zlib.error: Error -5 while decompressing data: incomplete or truncated stream`

When we changed https://github.com/kolide/fleet/blob/7153e0043c7430f46ffa8d766228def413eff56a/server/sso/authorization_request.go#L130 from `writer.Flush()` to `writer.Close()` the issue was resolved and we could no longer reproduce it.

```
fleet - version 2.0.2-9-g7153e00-dirty
  branch: 	master
  revision: 	7153e0043c7430f46ffa8d766228def413eff56a
  build date: 	2019-04-05T20:02:54Z
  build user: 	vagrant
  go version: 	go1.12.1
```

Ubuntu 16.04

### What did you do?

Configured SAML with Fleet. Attempted to do a service provider initiated auth.

I expect to see a SAML authnrequest come through that a SAML Identity Provider is able to successfully inflate.

### What did you see instead?

Upon Fleet sending a SAML AuthnRequest to a SAML IdP I get an error that it is unable to inflate the XML.

zwass:
Fixes an issue in which the requests are not able to be read by some SAML
request decoders.

Fixes #2024

### DIFF
--- a/server/sso/authorization_request.go
+++ b/server/sso/authorization_request.go
@@ -114,20 +114,26 @@ func getDestinationURL(settings *Settings) (string, error) {
 // See SAML Bindings http://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf
 // Section 3.4.4.1
 func deflate(xmlBuffer *bytes.Buffer) (string, error) {
+	// Gzip
 	var deflated bytes.Buffer
 	writer, err := flate.NewWriter(&deflated, flate.DefaultCompression)
 	if err != nil {
-		return "", err
+		return "", errors.Wrap(err, "create flate writer")
 	}
-	defer writer.Close()
 	n, err := writer.Write(xmlBuffer.Bytes())
 	if n != xmlBuffer.Len() {
+		_ = writer.Close()
 		return "", errors.New("incomplete write during compression")
 	}
 	if err != nil {
+		_ = writer.Close()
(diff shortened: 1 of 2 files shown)


zwass:
Verified I can still log in via GSuite SSO after this change.

PR #2026, opened 2019-04-09
Cleanup incoming hosts that do not complete enrollment

### LINKED ISSUE #1438: "Incoming Host" indefinitely when details are not retrieved after enrollment
In rare situations, a host may enroll, but shut down before returning distributed query results. In this case, we end up with "Incoming Host" forever.

Perhaps a cleanup job that removes enrolled hosts with no details every minute? If these are indeed legitimate hosts, they can re-enroll when they reconnect.

screen shot 2017-03-20 at 12 21 13 pm

zwass:
Deletes hosts that have enrolled but never updated their details (these
hosts show up as "incoming host" in the UI).

Closes #1438

### DIFF
--- a/cmd/fleet/serve.go
+++ b/cmd/fleet/serve.go
@@ -140,6 +140,7 @@ the way that the Fleet server works.
 				ticker := time.NewTicker(1 * time.Hour)
 				for {
 					ds.CleanupDistributedQueryCampaigns(time.Now())
+					ds.CleanupIncomingHosts(time.Now())
 					<-ticker.C
 				}
 			}()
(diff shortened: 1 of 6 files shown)


zwass:
Yeah, if a host is deleted and it still has a valid enroll secret it will just re-enroll. Under normal circumstances, hosts complete their details almost immediately after enrollment. I think 5 minutes gives a pretty generous margin. I'm not sure there is much benefit to exposing this as configuration and my recommendation would be to do so later if users request it.

blaedj:
Ok thanks, that answers the questions that I had!

PR #2061, opened 2019-06-11
Improve server and browser performance on host dashboard

zwass:
- Prevent a second (unnecessary) retrieval of all hosts when page loaded.
- Disable automatic reloading of hosts when over 100 hosts are present.

This is a stopgap solution to reduce the performance impact of
repeatedly loading many hosts. The final solution will be to implement
proper pagination.

### DIFF
--- a/frontend/pages/hosts/ManageHostsPage/ManageHostsPage.jsx
+++ b/frontend/pages/hosts/ManageHostsPage/ManageHostsPage.jsx
@@ -84,26 +84,9 @@ export class ManageHostsPage extends PureComponent {
   componentDidMount () {
     const { getEntities } = this;
     const { hosts } = this.props;
+    const silent = hosts.length > 0;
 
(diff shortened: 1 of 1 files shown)


frontend/pages/hosts/ManageHostsPage/ManageHostsPage.jsx
```diff
@@ -81,39 +81,33 @@ export class ManageHostsPage extends Component {
...
-    this.buildSortedHosts();
-
-    return this.getEntities();
-  }
-
   componentDidMount () {
     const { getEntities } = this;
+    const { hosts } = this.props;
 
     this.interval = global.window.setInterval(getEntities, 5000);
```
therealmik: Should this setInterval be replaced with a setTimeout on completion?

zwass: I took advantage of my current familiarity with this code to make this change.

PR #2079, opened 2019-07-25
Upgrade to Webpack 4

zwass:


zwass:
Looks like we will have to update the CI node version.

PR #2090, opened 2019-08-02
Use Github hosted assets in emails sent by Fleet

### LINKED ISSUE #1471: Email images should be served from an internet-accessible server
If users have their Kolide server configured to only be accessible behind their VPN, the images in emails sent from that Kolide instance may appear broken.

We could serve these images from a Kolide-owned server to ensure that the images are always accessible.

Thoughts?

zwass:
This change allows the images in Fleet emails to load properly from any
device with connectivity to github.com. Previously, emails might try to
load resources from a Kolide server not accessible from the email
client.

The asset URL will be based on the most recent git tag to accomodate
backwards-compatibility if the assets in the repo change.

Closes #1471

### DIFF
--- a/server/kolide/emails.go
+++ b/server/kolide/emails.go
@@ -49,7 +49,8 @@ type PasswordResetRequest struct {
 // SMTPTestMailer is used to build an email message that will be used as
 // a test message when testing SMTP configuration
 type SMTPTestMailer struct {
-	KolideServerURL string
+	KolideServerURL template.URL
+	AssetURL        template.URL
 }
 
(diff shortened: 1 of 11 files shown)


zwass:
IMO this doesn't need to be configurable. The new behavior should work well for almost all cases, and in cases it fails (no internet connectivity) the existing behavior wouldn't work either.

directionless:
There are certainly a handful of places that would not want to expose their use of fleet to github. But, I imagine we'll hear a feature request when/if it matters to them. 👍

Issue #2052, opened 2019-05-26
Feature Request - fleetctl export


TonyLeeVT:
Thanks for the great tool and support.

We are trying to share packs and queries with others, but are finding it a bit tedious to export each pack and query manually.  It would be fantastic if we had a command such as "export" that would provide the following functionality:

fleetctl export p - export all packs
fleetctl export q - export all queries
fleetctl export l - export all labels
fleetctl export p q - export packs and queries
fleetctl export p "<name>" - export specified pack and all associated queries

All output would be able to be imported using the fleetctl apply command.

I hope this covers all of the scenarios.  If I missed anything that would be useful, please feel free to add it.

Issue #1975, opened 2019-01-03
Add option to get queries along with pack in fleetctl get -p


benbasscom:
### What version of `fleet` are you using (`fleet version --full`)?
{
  "version": "2.0.1",
  "branch": "master",
  "revision": "d01727dff26683ba29424fbdb7e58e403ab6210d",
  "go_version": "go1.11.2",
  "build_date": "2018-11-26T19:21:48Z",
  "build_user": "zwass"
}
### What operating system are you using?
RHEL 7.6 server, macOS 14.2 client/admin

### What did you do?
fleetctl get -p "my awesome pack"

### What did you expect to see?

Yaml output of the entire pack, including the queries, enabling applying the pack to generate a full pack in fleet.  This would match the output of converting a json pack to yaml.

### What did you see instead?
Yaml output of the pack without the queries at the bottom, just the names and schedules of the queries.

zwass:
Technically a pack definition does not include the query SQL (because queries may be shared across multiple packs), but I can see how this would be a useful feature.

TonyLeeVT:
It appears that the request is to add an option that does the following:
1)  First retrieve the pack
2)  For each query listed in the pack, retrieve the query

This provides a bit of an export feature from fleetctl which is also what I am seeking.  This is useful for backing up a fleet's packs and queries as well as making it easier to package and share packs and queries with others.  Thanks for your continued support.

icebawkz:
A dirty workaround for this can be achieved using pipes:

`fleetctl get pack unwanted-chrome-extensions && fleetctl get pack unwanted-chrome-extensions | grep -Po "(?<=name: )[^\n]*" | xargs -n1 -I {} sh -c "fleetctl get query {}"`

```
apiVersion: v1
kind: pack
spec:
  id: 4
  name: unwanted-chrome-extensions
  queries:
  - description: (https://www.bleepingcomputer.com/news/security/eight-chrome-extensions-hijacked-to-deliver-malicious-code-to-4-8-million-users/)
    interval: 3600
    name: TouchVPN
    query: TouchVPN
  - description: (https://www.bleepingcomputer.com/news/security/chrome-extension-with-over-one-million-users-hijacked-to-serve-adware/)
    interval: 3600
    name: WebDeveloper
    query: WebDeveloper
  - description: (https://www.virustotal.com/#/file/5cab0821f597100dc1170bfef704d8cebaf67743e9d509e83b0b208eb630d992/detection)
    interval: 3600
[...]

PR #2092, opened 2019-08-09
fleetctl: add --yaml flag to get command for labels, packs, and queries

therealmik:
Add the `--dump` flag to the `get` subcommand for packs and queries, for use with `fleetctl apply` on another system (for example, moving from staging to prod).

### DIFF
--- a/cmd/fleetctl/get.go
+++ b/cmd/fleetctl/get.go
@@ -11,6 +11,10 @@ import (
 	"github.com/urfave/cli"
 )
 
+const (
+	yamlFlagName = "yaml"
+)
+
 type specGeneric struct {
 	Kind    string      `json:"kind"`
 	Version string      `json:"apiVersion"`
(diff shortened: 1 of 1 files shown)


therealmik:
> Nice, thank you. Now we just need to add a flag for including the queries as well when getting a pack.

@zwass I considered implementing a `get all-config` command, but you can get everything you need from the `get labels`, `get queries` and `get packs` commands separately, and `get options` already did yaml output.

I'm pretty sure this is all you need to convert to managed config files per https://github.com/kolide/fleet/blob/master/docs/cli/file-format.md

zwass:
Yes, agreed if you want to export the _entire_ configuration. However I think folks are interested in being able to export a single pack along with all of the queries in that pack.

See https://github.com/kolide/fleet/issues/2052 and https://github.com/kolide/fleet/issues/1975.

PR #2093, opened 2019-08-11
Allow scheduled query shard to be set to empty in frontend

### LINKED ISSUE #1452: Error updating packs
Changing the shard number to an empty value causes a 500 error.

zwass:
Fixes #1452

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -271,6 +271,14 @@
   version = "v1.2.0"
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


zwass:
IMO we really ought to move towards an API where the frontend sends the entire updated object. I think there's a lot of potential tricky situations with trying to use the update style where the frontend only sends the values to be updated, and I don't believe the small savings in network traffic is worth it.

Nevertheless, this does fix the issue without requiring a major refactor.

PR #2108, opened 2019-09-21
Rename remaining file in Admin to admin

directionless:
We have files spread across both `Admin` and `admin`. On an osx machine, those are the same directory, which has probably masked some underlying errors. This normalizes the case to the one that has more.

I did not test this, it's possible we should make this all title case instead.


directionless:
If you can easily test, I'd say merge. I don't have a good test bed for fleet.

zwass:
I gave it a manual test by building Fleet and loading up the page and it seems fine to me.

PR #2112, opened 2019-09-24
Add ability to prefix Fleet URLs

### LINKED ISSUE #1661: Path prefix for web interface URLs
I'm running the Fleet management interface behind a reverse proxy.
I would like to access it by specific path, for example "/apps/fleet".

Currently, the interface redirects to paths such as "/login" and "/hosts", which breaks this logic.
It would be nice if there was a configuration option to specify a "path prefix" to be added to URLs in the JavaScript code and redirects.

zwass:
- Add the server_url_prefix flag for configuring this functionality
- Add prefix handling to the server routes
- Refactor JS to use appropriate paths from modules
- Use JS template to get URL prefix into JS environment
- Update webpack config to support prefixing

Closes #1661


frontend/components/side_panels/SiteNavSidePanel/navItems.js
```diff
@@ -1,27 +1,30 @@
...
+
 export default (admin) => {
   const adminNavItems = [
     {
       icon: 'admin',
       name: 'Admin',
       location: {
-        regex: /^\/admin/,
-        pathname: '/admin/users',
+        regex: new RegExp(`^${URL_PREFIX}/admin/`),
```
directionless, line 10: This, and more, break if URL_PREFIX has a trailing slash.

I think thats enforced by the regex, but it feels weirdly fragile to me

frontend/router/index.jsx
```diff
@@ -27,13 +27,15 @@ import Kolide404 from 'pages/Kolide404';
...
+import PATHS from 'router/paths';
+import URL_PREFIX from 'router/url_prefix';
 
 const history = syncHistoryWithStore(browserHistory, store);
 
 const routes = (
   <Provider store={store}>
     <Router history={history}>
-      <Route path="/" component={App}>
+      <Route path={`${URL_PREFIX}/`} component={App}>
```
directionless: Maybe silly, but why isn't this is in PATHs

frontend/components/LoginRoutes/LoginRoutes.jsx
```diff
@@ -60,8 +60,8 @@ const mapStateToProps = (state, ownProps) => {
   const { location: { pathname, query } } = ownProps;
   const { token } = query;
 
-  const isForgotPassPage = pathname === '/login/forgot';
-  const isResetPassPage = pathname === '/login/reset';
+  const isForgotPassPage = pathname.endsWith('/login/forgot');
+  const isResetPassPage = pathname.endsWith('/login/reset');
```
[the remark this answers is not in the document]
zwass: Nothing is gated behind this. The user can read the whole source of the app including for these pages when they load `bundle.js`, or by coming to this repository. What matters is the data sent over by the API, and those requests are always authenticated on the backend.

frontend/components/side_panels/SiteNavSidePanel/navItems.js
zwass: Do you have some idea of how else we could handle this besides at the entry point of the program? I would not like to enforce or strip this at each point it is used.

cmd/fleet/serve.go
```diff
@@ -262,13 +273,19 @@ the way that the Fleet server works.
 				if err != nil {
 					initFatal(err, "generating debug token")
 				}
-				r.Handle("/debug/", http.StripPrefix("/debug/", netbug.AuthHandler(debugToken)))
+				rootMux.Handle("/debug/", http.StripPrefix("/debug/", netbug.AuthHandler(debugToken)))
 				fmt.Printf("*** Debug mode enabled ***\nAccess the debug endpoints at /debug/?token=%s\n", url.QueryEscape(debugToken))
 			}
 
+			if len(config.Server.URLPrefix) > 0 {
```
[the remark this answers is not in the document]
zwass: Do you mean something besides at the entry point with the regex?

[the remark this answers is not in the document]
directionless: If I understand this all correctly, right now we enforce `"^(?:/[a-zA-Z0-9_.~-]+)+$"`. Which says "any number of `/[a-z]` blobs. Notable, this means we require the leading `/` and we deny a trailing `/`. 

I assume this is so that future usage is very easy -- we can add it to the path, and we know the slashes will work out correctly.

However, I find this to be somewhat cumbersome from an end user perspective. I often feel unsure whether or not which of the 4 slash options I use. Leading? Trailing? Both? None? I often just have to try it and see what the software accepts.

I think that as written this is correct -- the regex enforces the underlying code requirements.

An alternative would be to add some rewriting at the entry point. Whether or not this is a good idea depends where you fall on the "explicit" vs "forgiving" axis. But I could imagine some code like:
```
config.Server.URLPrefix = strings.TrimSuffix(config.Server.URLPrefix, "/")
if len(config.Server.URLPrefix) > 0 {
  if !strings.HasPrefix(config.Server.URLPrefix, "/") {
    config.Server.URLPrefix = "/" + config.Server.URLPrefix
  }
  if !allowedURLPrefixRegexp.MatchString(config.Server.URLPrefix) {
   ...
}
}
```

I think this is a primarily a user convenience method

frontend/components/side_panels/SiteNavSidePanel/navItems.js
directionless: See above.

PR #2116, opened 2019-10-10
Allow host search by UUID

piax93:
When searching hosts, querying the fleet using hostnames and/or IPs may lead to inconsistent results since these can easily vary in time. On the other hand, UUIDs should be a much more "static" way to identify a device.
This change implements search by UUID for the MySQL and in memory datastores.


piax93:
I just realized I need to update the MySQL schema, working on it.

zwass:
Please ping when you have updated this and are ready for review :)

piax93:
Hi @zwass, I already did in 3a30930, it's ready to be reviewed, sorry for forgetting to comment.

PR #2117, opened 2019-10-11
Allow the ability to add automatic host expiry in Fleet admin configu…

### LINKED ISSUE #1860: Allow more aggressive cleanup of hosts
When using fleet in an environment that consists of a large number of ephemeral hosts the number of "offline" hosts can grow considerably, leading to unexpected behavior.

In environments like our own with disposable VMs and where there's a considerable difference between the peak and trough numbers of live hosts the "offline" status for hosts serves no function as once a host goes "offline" (is terminated) it will never reappear online.

Ideally it'd be possible to set more aggressive cleanup intervals for hosts by platform, as were we to start using this same deployment of fleet with our OS X endpoints *which do go offline* we'd obviously prefer not to have these temporarily offline hosts removed from fleet.

austburn:
…ration.

Fixes #1860

Screenshot_2019-10-11 Kolide Fleet

### DIFF
--- a/server/datastore/mysql/app_configs.go
+++ b/server/datastore/mysql/app_configs.go
@@ -31,7 +31,7 @@ func (d *Datastore) isEventSchedulerEnabled() (bool, error) {
 		return false, err
 	}
 	if !rows.Next() {
-		return false, errors.New("Error detecting event scheduler status.")
+		return false, errors.New("Error detecting MySQL event scheduler status.")
 	}
 	var value string
 	if err := rows.Scan(&value); err != nil {
@@ -58,7 +58,7 @@ func (d *Datastore) SaveAppConfig(info *kolide.AppConfig) error {
 	}
 
 	if !eventSchedulerEnabled && info.HostExpiryEnabled {
-		return errors.New("Event Scheduler must be enabled for configuring Host Expiry.")
+		return errors.New("MySQL Event Scheduler must be enabled to configure Host Expiry.")
 	}
 
 	if err := d.ManageHostExpiryEvent(info.HostExpiryEnabled, info.HostExpiryWindow); err != nil {
--- a/server/datastore/mysql/app_configs.go
+++ b/server/datastore/mysql/app_configs.go
@@ -1,8 +1,14 @@
 package mysql
 
 import (
+	"fmt"
+
 	"github.com/kolide/fleet/server/kolide"
 	"github.com/pkg/errors"
+)
+
+const (
+	HostExpiryEvent = "host_expiry"
 )
 
 func (d *Datastore) NewAppConfig(info *kolide.AppConfig) (*kolide.AppConfig, error) {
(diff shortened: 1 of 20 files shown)


server/datastore/mysql/app_configs.go
```diff
@@ -23,6 +29,32 @@ func (d *Datastore) AppConfig() (*kolide.AppConfig, error) {
...
+	var value string
+	if err := rows.Scan(&value); err != nil {
+		return err
+	}
+	if value != "ON" {
+		return errors.New("Event Scheduler must be enabled for configuring Host Expiry.")
+	}
+
+	if !hostExpiryEnabled {
+		_, err := d.db.Exec(fmt.Sprintf("DROP EVENT IF EXISTS %s", HostExpiryEvent))
```
zwass: We avoid using string formatting to craft SQL queries in order to protect against SQL injection attacks. Please use parameters as you can see done in https://github.com/kolide/fleet/blob/master/server/datastore/mysql/app_configs.go#L27-L114.

server/datastore/mysql/app_configs.go
```diff
@@ -23,6 +29,32 @@ func (d *Datastore) AppConfig() (*kolide.AppConfig, error) {
...
+		return err
+	}
+
+	_, err = d.db.Exec(
+		fmt.Sprintf("CREATE EVENT IF NOT EXISTS %s ON SCHEDULE EVERY 1 HOUR ON COMPLETION PRESERVE DO DELETE FROM HOSTS WHERE DATE_ADD(day, -%d, NOW()) < updated_at", HostExpiryEvent, hostExpiryWindow))
```
zwass: Please use parameters here as well.

zwass: A couple things:

1. Prefer to use `seen_time` for the last time the host checked in.

2. Does this `DATE_ADD` syntax work for you? I get
```
Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-2, NOW()) < updated_at' at line 1
```
It looks to me like the correct syntax (based on https://dev.mysql.com/doc/refman/5.7/en/date-and-time-functions.html#function_date-add) might be `WHERE seen_time < DATE_SUB(NOW(), INTERVAL 2 DAYS)` (reversing the comparison and using `DATE_SUB` makes it easier for me to read).

server/datastore/mysql/app_configs.go
```diff
@@ -23,6 +29,32 @@ func (d *Datastore) AppConfig() (*kolide.AppConfig, error) {
...
+	}
+	var value string
+	if err := rows.Scan(&value); err != nil {
+		return err
+	}
+	if value != "ON" {
+		return errors.New("Event Scheduler must be enabled for configuring Host Expiry.")
+	}
+
+	if !hostExpiryEnabled {
```
zwass, line 45: Definitely an edge case, but it could be good to move this above the check for the event scheduler so that a user can disable this setting after the event scheduler has been turned off on their MySQL instance.

server/datastore/mysql/app_configs.go
austburn: Ah, cool, thank you! My SQL is shoddy at best :+1:

austburn: Turns out that `CREATE EVENT` is not a supported prepared statement. I've just added the event name in the statements (i.e. `DROP EVENT IF EXISTS host_expiry`), but I can't seem to get around the `INTERVAL ? DAY` bit - thoughts on using `fmt.Sprintf` just for the string substitution of the integer window value?

austburn: https://dev.mysql.com/doc/refman/8.0/en/sql-syntax-prepared-statements.html - See "SQL Syntax Allowed in Prepared Statements"

zwass: Well that is appalling that prepared statements don't support creating events. We do have a couple mitigations here:

1) The user-controlled input is formatted as an integer in the `fmt.Sprintf`.
2) Only admins may modify the configuration value.

austburn:
opted to pull the event scheduler status check into a function for readability and do the checks prior to saving the config

PR #2121, opened 2019-10-21
Allow host expiry window to be altered

### LINKED ISSUE #2122: Host Expiry Window is not updated when changed but host expiry left enabled
`master`

### What did you do?
In the admin console, enable host expiry and set a window. Leaving host expiry enabled, change the window to a different amount of days. The setting is saved, but the host expiry window is not actually altered in the query.

The query should be updated to remove hosts that haven't checked in the newly set number of days.

### What did you see instead?
The setting itself is saved, but the query is not updated.

austburn:
Bug on my behalf - if you had left host expiry enabled but changed the window, the event would not be altered to reflect that. Event creation doesn't support upserts unfortunately. SQL is not my forte, so let me know if there's a better way to accomplish this.

Fixes #2122

### DIFF
--- a/server/datastore/mysql/app_configs.go
+++ b/server/datastore/mysql/app_configs.go
@@ -43,12 +43,24 @@ func (d *Datastore) isEventSchedulerEnabled() (bool, error) {
 
 func (d *Datastore) ManageHostExpiryEvent(hostExpiryEnabled bool, hostExpiryWindow int) error {
 	var err error
-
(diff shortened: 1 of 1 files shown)


austburn:
I tried to make this as easy to read as possible as the logic gets a little convoluted.

zwass:
Hey @austburn I simplified the logic a bit and I'm ready to merge, but would love to have you take a look at the changes before I do so. Thanks!

austburn:
:shipit: Thanks! I think I looked at this too long :stuck_out_tongue_closed_eyes:

PR #2123, opened 2019-10-30
Fix #2021: Allow users to be deleted

### LINKED ISSUE #2021: Allow users to be deleted
Currently, users can be disabled (which logs them out and prevents them from logging back in), but cannot be deleted. It makes sense to add an action to permanently delete users.

nyanshak:
* Adds API endpoints to allow deleting users. Following patterns in the
rest of the API, to delete one user, use `DELETE
/api/v1/kolide/users/{id}" and to delete multiple users, `POST` an array
of ids to `/api/v1/kolide/users/delete`.


zwass:
Thanks for contributing this, and apologies for the delay in review. I'd like to get user deletion merged, but I am concerned about using soft-deletion as the strategy. Ideally I'd like to address https://github.com/kolide/fleet/issues/2146 and then hard-delete users.  For now we have "disable" which is fairly similar to soft deletion.

Do you have a specific use case for this right now? If not I'd like to hold off on merging until we can address #2146 and then try to do this more thoroughly including support in the frontend and/or fleetctl.

nyanshak:
I agree with that approach (hard-delete vs soft-delete). I was trying to follow the pattern I saw in the other tables, but was hoping to raise that as a separate issue later :)

zwass:
We can revisit this now that #2327 has been merged.

PR #2137, opened 2019-11-07
create a dockerized kolide experience

kavehmz:
This change convert the docker-compose so we can test it only with docker as dependency.
no need to have node/go other other tools locally.

```bash
docker-compose up
```
When all containers start eventually (they can fail and restart) we can open https://localhost and login with:
`username: kolide@kolide.com`
`password: kolide`

There will be already one node registered there.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       KOLIDE_MYSQL_ADDRESS: mysql:3306
       KOLIDE_MYSQL_PASSWORD: kolide
       KOLIDE_REDIS_ADDRESS: redis:6379
-      KOLIDE_AUTH_JWT_KEY: E6VMWGBgZkKzfY/dw3zsZMd9foPsUrmL
+      KOLIDE_AUTH_JWT_KEY: FAKESECRETFake/thisisjustfakekey
     ports:
       - "443:443"
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
       - "(
           while ! grep -q 00:01BB /proc/net/tcp6;do echo wait;sleep 1;done; \
           fleetctl config set --address https://localhost:443 --tls-skip-verify; \
-          fleetctl setup --email kolide@kolide.com --password kolide; \
+          fleetctl setup --email kolide@example.com --password kolide; \
           fleetctl get enroll-secret > /data/enroll-secret;cp ./tools/osquery/kolide.crt /data/ \
         ) & \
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -1,9 +1,36 @@
...
+          fleetctl setup --email kolide@kolide.com --password kolide; \
+          fleetctl get enroll-secret > /data/enroll-secret;cp ./tools/osquery/kolide.crt /data/ \
+        ) & \
+        fleet prepare db && fleet serve"
+    environment:
+      KOLIDE_SERVER_ADDRESS: 0.0.0.0:443
+      KOLIDE_MYSQL_ADDRESS: mysql:3306
+      KOLIDE_MYSQL_PASSWORD: kolide
+      KOLIDE_REDIS_ADDRESS: redis:6379
+      KOLIDE_AUTH_JWT_KEY: E6VMWGBgZkKzfY/dw3zsZMd9foPsUrmL
```
directionless: Is there any significance to this? If it's a random string, can it start "FAKESECRET" or something to be clear?

docker-compose.yml
```diff
@@ -1,9 +1,36 @@
...
+      dockerfile: Dockerfile.compose
+    image: kolide-local
+    command:
+      - /bin/bash
+      - "-c"
+      - "(
+          while ! grep -q 00:01BB /proc/net/tcp6;do echo wait;sleep 1;done; \
+          fleetctl config set --address https://localhost:443 --tls-skip-verify; \
+          fleetctl setup --email kolide@kolide.com --password kolide; \
+          fleetctl get enroll-secret > /data/enroll-secret;cp ./tools/osquery/kolide.crt /data/ \
```
directionless, line 15: <3 this

docker-compose.yml
kavehmz: Good point. I changed it to one that shows clearly it is a fake key.

docker-compose.yml
```diff
@@ -1,9 +1,36 @@
...
+      context: .
+      dockerfile: Dockerfile.compose
+    image: kolide-local
+    command:
+      - /bin/bash
+      - "-c"
+      - "(
+          while ! grep -q 00:01BB /proc/net/tcp6;do echo wait;sleep 1;done; \
+          fleetctl config set --address https://localhost:443 --tls-skip-verify; \
+          fleetctl setup --email kolide@kolide.com --password kolide; \
```
directionless: If this is an example, it should be in `example.com` not kolide.com
```suggestion
          fleetctl setup --email kolide@example.com --password kolide; \
```

PR #2142, opened 2019-11-08
Remove Support for Deprecated TLSProfileOld

alrs:
The Go authors deprecated support for SSLv3 in go 1.13. They've decided to remove it in go 1.14, and have already merged the commit. 

See: https://golang.org/issue/32716

This PR, or something like it, should probably land before February 2020.


alrs:
> What do you think of changing the minimum version to drop SSL3 support without removing the "old" profile entirely?

It's tricky, yeah? If someone is depending on the behavior of "old" and it changes, is that better than having "old" disappear entirely, break something for someone, but make the change obvious?

zwass:
I added a deprecation warning for this in #2183 and I think we can merge the removal after the next release goes out with that warning.

jalseth:
@zwass Can we merge this? The deprecation was introduced in 2.5.0 which has been out for over 6 months.

Issue #1953, opened 2018-11-12
Add fleetctl support for Admin Tab.


stefanmaerz:
Creating an Issue per my conversation with @zwass on osquery's Slack channel

It would be awesome if I could manage the "Admin" tab (specifically App Settings) via fleetctl/yaml files. This  would help Fleet admins keep the configuration under revision control and help make deploying fleet even more reproducible.

Best Regards,
-Stefan

zwass:
This was addressed by #2143.

PR #2143, opened 2019-11-08
add fleetctl get/apply config

billcobbler:
Implements #1953


billcobbler:
I'll update the PR with docs to address 2. 

For 1, I'm not sure. As is, the get returns the masked *****. That's an attempt to dissuade storing creds I think. We could check for the masked value and ignore it on the modify operation. So if a user exports and then imports they don't overwrite their SMTP password with *****.

PR #2153, opened 2019-11-13
Harden Docker container

hazcod:
* produce go static builds
* remove unused libraries
* remove shell

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM alpine
-MAINTAINER Kolide Developers <engineering@kolide.co>
-
-RUN apk --update add ca-certificates
+FROM gcr.io/distroless/base-debian10:nonroot
+LABEL author Kolide Developers <engineering@kolide.co>
+USER nonroot
 
 COPY ./build/binary-bundle/linux/fleet ./build/binary-bundle/linux/fleetctl /usr/bin/
 
-CMD ["fleet", "serve"]
+EXPOSE 8080
+CMD ["/usr/bin/fleet", "serve"]
--- a/Makefile
+++ b/Makefile
@@ -107,10 +107,10 @@ endif
 build: fleet fleetctl
 
 fleet: .prefix .pre-build .pre-fleet
-	go build -i -o build/${OUTPUT} -ldflags ${KIT_VERSION} ./cmd/fleet
+	go build -i -o build/${OUTPUT} -ldflags ${KIT_VERSION} -ldflags "-w -s -extldflags '-static'" ./cmd/fleet
 
 fleetctl: .prefix .pre-build .pre-fleetctl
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM alpine
-MAINTAINER Kolide Developers <engineering@kolide.co>
-
-RUN apk --update add ca-certificates
+FROM gcr.io/distroless/base-debian10:nonroot
+LABEL author Kolide Developers <engineering@kolide.co>
```
zwass: Let's use
```
LABEL maintainer="engineering@kolide.co"
```
as suggested in https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

Makefile
```diff
@@ -107,10 +107,10 @@ endif
 build: fleet fleetctl
 
 fleet: .prefix .pre-build .pre-fleet
-	go build -i -o build/${OUTPUT} -ldflags ${KIT_VERSION} ./cmd/fleet
+	go build -i -o build/${OUTPUT} -ldflags ${KIT_VERSION} -ldflags "-w -s -extldflags '-static'" ./cmd/fleet
```
zwass, line 110: Is this change necessary when using `base-debian10`? https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md seems to suggest that this may not be needed.

hazcod: Exactly right, I've changed it to `static` since it now receives glibc from the original go image used to built the code from `.circleci/`. Seems clearer

zwass: Sorry it wasn't clear, but my preferred strategy would be to leave the build the same while using the `base-debian10` image. Do you see a compelling reason to change the build flags for all binaries vs. using the base image?

hazcod: @zwass care to share your relationale behind that? Since distroless does **not** have an update policy, there is no guarantee when they pull in new glibc versions so it would make more sense to me to package that in during compilation. https://github.com/GoogleContainerTools/distroless/issues/326

zwass: My rationale is that we are changing the build flags for all the binaries produced by this Makefile, not just those being placed in the Docker container. Maybe we can refactor the Makefile in such a way to enable the fully static build for the binary used in the container while leaving other binary builds unchanged?

hazcod: I think it would make more sense to build fleet in a multi-stage container, what do you think? So we can move `GOOS=linux go build -i -o build/linux/${OUTPUT} -ldflags ${KIT_VERSION} ./cmd/fleet` to the build container.

zwass: I am on board with that strategy. It will likely require running `make deps-js && make generate-js` in a Node container, then `make deps-go && make generate-go && make` in a Go container.

PR #2159, opened 2019-12-03
check if status is not nil before deciding that all hosts have responded

hectornvm:
Currently, the `total == responded` line in `query.go` will resolve to true if the Fleet server has not returned any data, as 0/0 hosts are returning any information. However, this may not necessarily mean that Fleet should exit, as the values also resolve to zero if the websocket responses have not been return yet. Forcing status to also be not nil allows us to keep the connection ongoing until responses are returned.
This contribution was co-researched with @4ndygu

### DIFF
--- a/cmd/fleetctl/query.go
+++ b/cmd/fleetctl/query.go
@@ -160,7 +160,7 @@ func queryCommand() cli.Command {
 					if !flQuiet {
 						s.Suffix = msg
 					}
-					if total == responded {
+					if status != nil && total == responded {
 						s.Stop()
 						if !flQuiet {
 							fmt.Fprintln(os.Stderr, msg)


4ndygu:
Hi @zwass! Is it possible for us to submit under an Apache 2.0 license? We're asking as this is committed under our work identities, and it may be easier for us to commit using that rather than getting legal to review and our lead to approve the CLA. If not, we can move forward with that other process :).

zwass:
The project requires signing the CLA in order to accept submissions. I know this can be an onerous process and we appreciate you taking the time to get it approved.

zwass:
Due to pending CLA, I am replacing this with #2184.

Issue #2140, opened 2019-11-08
Allow disabling ad-hoc querying panel


nyanshak:
Feature request: support disabling distributed queries through API / query panel page.

Fleet internally uses distributed querying to get info such as network interfaces. I'd like to be able to leave this enabled, but disable the distributed querying features of fleet, such that users are not able to go to the query page and run queries (or use the API).

It's possible to set `disable_distributed` to true, but then hosts won't send info that fleet uses to enrich the UI.

Also, when `disable_distributed` is set to true, the UI still lets you go to the query page, select `All Hosts` and queue up an ad-hoc query that won't ever return results. It would be better for users to remove the link from the sidebar or maybe show a warning that the feature is disabled when that is set to disabled.

anelshaer:
I was thinking also disabling ad-hoc queries only admins can write queries, and normal users only use already created/saved queries made by admin to reduce the risk of running something that causes production issues.

nyanshak:
I think that would be great, as well as limiting those users to being able to query certain numbers of hosts. However, I think it's a bit off-topic for this ticket, and probably belongs instead in https://github.com/kolide/fleet/issues/2077

zwass:
This has been implemented by #2167.

mbigras:
Does disabling live queries remove the dependency on redis?

zwass:
Redis is also used for SSO login. You could remove Redis and you would find that most things work but `healthz` returns 500 errors and SSO login does not work.

PR #2167, opened 2019-12-24
add ability to disable live queries

billcobbler:
Implements #2140 
- added toggle to disable live queries in advanced settings
- added new live query status endpoint (checks for disabled via config and Redis health)
- updated QueryPage UI to use new live query status endpoint


anelshaer:
is there ability to enable it for specific users, or its only enable/disable for all users.

billcobbler:
@anelshaer, it's just a simple global toggle and applies to all users but a sneaky admin user could toggle it off/on :smile:. 

Should be rather simple to add a bypass for admin users, but targeting specific users would be a larger change (I think) because I believe the only user segmentation that exists currently is admin/non-admin.

server/datastore/mysql/migrations/tables/20191220130734_AddLiveQueryDisabledToAppConfig.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func init() {
+	MigrationClient.AddMigration(Up20191220130734, Down20191220130734)
+}
+
+func Up20191220130734(tx *sql.Tx) error {
+	_, err := tx.Exec(
+		"ALTER TABLE `app_configs` " +
+			"ADD COLUMN `live_query_disabled` TINYINT(1) NOT NULL DEFAULT FALSE;",
```
groob, line 14: I suggest using a boolean here to improve readability.

server/kolide/app.go
```diff
@@ -228,8 +231,9 @@ type OrgInfo struct {
 
 // ServerSettings contains general settings about the kolide App.
 type ServerSettings struct {
-	KolideServerURL *string `json:"kolide_server_url,omitempty"`
-	EnrollSecret    *string `json:"osquery_enroll_secret,omitempty"`
+	KolideServerURL    *string `json:"kolide_server_url,omitempty"`
+	EnrollSecret       *string `json:"osquery_enroll_secret,omitempty"`
+	LiveQueryDisabled *bool   `json:"live_query_disabled,omitempty"`
```
groob: gofmt?

server/datastore/mysql/migrations/tables/20191220130734_AddLiveQueryDisabledToAppConfig.go
billcobbler: I was being consistent with the existing code base in that every other setting flag uses TINYINT(1) already.

server/kolide/app.go
billcobbler: Fixed.

zwass:
Follow up in #2181.

PR #2176, opened 2020-01-09
Fix error initializing Fleet when MySQL EVENT permissions missing

### LINKED ISSUE #2174: Disable host_expiry_enabled on setup and provide cli command to run the cleanup
2.4.0

Alpine

### What did you do?
Upon setting up a new kolide instance, when connected to a database that does not provide `EVENT` access, kolide fails to initialize.

`/api/v1/setup` returns a 500 and we are unable to proceed with the setup because the host_expiry event cannot be dropped because we don't have access to it.

```json
{
  "message": "Unknown Error",
  "errors": [
    {
      "name": "base",
      "reason": "new app config: drop existing host_expiry event: Error 1044: Access denied for user 'kolide'@'%' to database 'kolide_fleet'"
    }
  ]
}
```

Setup to proceed and notify the user event has been disabled due to lack of permissions.

### What did you see instead?

Stuck on setup page until going into the database and editing the `app_configs` table and setting `host_expiry_enabled` to `1` (false).

For those of us who are using a managed database that does not provide access to `EVENT` or has the event scheduler disabled, it would be nice if we could instead run a fleet command to do the host cleanup and then we can schedule it with whatever mechanism we want.

zwass:
Fixes #2174

### DIFF
--- a/server/datastore/mysql/app_configs.go
+++ b/server/datastore/mysql/app_configs.go
@@ -3,6 +3,8 @@ package mysql
 import (
 	"fmt"
 
+	"github.com/VividCortex/mysqlerr"
+	"github.com/go-sql-driver/mysql"
 	"github.com/kolide/fleet/server/kolide"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


zwass:
The other lib is an existing dependency, though it just contains constants mapping to the MySQL error codes. We could certainly eliminate it at some point by putting in our own constants in the `mysql` package.

PR #2186, opened 2020-01-17
Implement fleetctl goquery

zwass:
- Update fleet APIs to support necessary operations in goquery
- Implement support for goquery in fleetctl


cmd/fleetctl/goquery.go
```diff
@@ -0,0 +1,132 @@
...
+type activeQuery struct {
+	status  string
+	results []map[string]string
+}
+
+type goqueryClient struct {
+	client       *service.Client
+	queryCounter int
+	queries      map[string]activeQuery
+	// goquery passes the UUID, while we need the hostname (or ID) to
```
directionless, line 26: I don't know if this is totally right? goquery passes whatever you handed it. So you could hand it a hostname+query id parameter, and it'll pass them back

zwass: As of now goquery uses the UUID of the returned host object to refer to the host later. I discussed with those folks the possibility of goquery using the original identifier, but that has not yet been implemented.

directionless: Yes, it's reused. But, IIRC, it can be whatever you want it to be. A ulid is just a string

Issue #2133, opened 2019-11-06
log x-forwarded-for IP


groob:
If fleet is behind a loadbalancer, the logs for `ip_addr` will display the ip of the proxy, not the remote ip of the client. 

Fleet should also log the remote IP from the x-forwarded-for field.

To accomplish this, all the log entries will need an additional line that reads `ContextKeyRequestXForwardedFor`

Example: 

https://github.com/kolide/fleet/blob/d453ca3f8f2db7c2b73fcf04f4ac85c05084be64/server/service/logging_osquery.go#L32-L49

Becomes 

	defer func(begin time.Time) {
		_ = mw.logger.Log(
			"method", "AuthenticateHost",
			"ip_addr", ctx.Value(kithttp.ContextKeyRequestRemoteAddr).(string),
			"x_for_ip_addr", ctx.Value(kithttp.ContextKeyRequestXForwardedFor).(string),
			"err", err,
			"took", time.Since(begin),
		)
	}(time.Now())

zwass:
Closed by #2193.

PR #2193, opened 2020-02-10
Added support to logging X-Forward-For header (#2133)

jemos:
This patch is just the code as suggested by @groob in issue #2133. It was tested and is working both when the header is present and when it's not (x_for_ip_addr is then set to "" in the logs.)


zwass:
Nice, thank you! I imagine we may want to do similar for the other non-osquery endpoints, but I'll be happy to merge this if you can get the CLA signed.

jemos:
My commit had the wrong email address so I wasn't able to pass the license/cla. The only way I found to fix it was to force push an amend =/

Regarding the other endpoints I agree, I will try to do it later.

zwass:
Amending the commit works just fine, thank you!

Issue #845, opened 2017-01-10
Getting lots of mysql deadlock errors submitting distributed query results


groob:
The issue appears to be coming from here: https://github.com/kolide/kolide-ose/blob/ee3940e1638834ad737fe367c865d79645953aee/server/datastore/mysql/hosts.go#L213-L217

```
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=963.841818ms
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=629.534269ms
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=956.276357ms
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=3.995755211s
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=898.475581ms
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=470.400522ms
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=344.661257ms
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=475.22221ms
ts=2017-01-10T06:32:16Z component=service method=SubmitDistributedQueryResults err="failed to update host details: updating nics: Error 1213: Deadlock found when trying to get lock; try restarting transaction" took=480.896582ms
[...]

PR #2203, opened 2020-03-04
retry sqlx transactions

billcobbler:
An attempt at addressing #845 by adding an exponential retries to methods that use sqlx.Tx.

### DIFF
--- a/server/datastore/mysql/mysql.go
+++ b/server/datastore/mysql/mysql.go
@@ -61,23 +61,24 @@ func (d *Datastore) withRetryTxx(fn txFn) (err error) {
 			return errors.Wrap(err, "creating transaction")
 		}
 
-		err = fn(tx)
-
 		defer func() {
 			if p := recover(); p != nil {
 				tx.Rollback()
 				panic(p)
-			} else if err != nil {
-				rbErr := tx.Rollback()
-				if rbErr != nil && rbErr != sql.ErrTxDone {
-					panic(fmt.Sprintf("got err '%s' rolling back after err '%s'", rbErr, err))
-				}
-			} else {
-				err = tx.Commit()
 			}
 		}()
 
-		return err
+		err = fn(tx)
+		if err != nil {
+			rbErr := tx.Rollback()
+			if rbErr != nil && rbErr != sql.ErrTxDone {
+				panic(fmt.Sprintf("got err '%s' rolling back after err '%s'", rbErr, err))
+			}
+		} else {
+			err = tx.Commit()
+		}
+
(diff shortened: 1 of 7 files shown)


zwass:
This seems like a good idea. Have you tested it under a real workload?

billcobbler:
Not our real workload, but testing locally with 100 osquery containers was enough to recreate the deadlocks seen in #845, and this worked in that scenario.

server/datastore/mysql/mysql.go
```diff
@@ -50,6 +51,40 @@ func (d *Datastore) getTransaction(opts []kolide.OptionalArg) dbfunctions {
...
+				if rbErr != nil && rbErr != sql.ErrTxDone {
+					panic(fmt.Sprintf("got err '%s' rolling back after err '%s'", rbErr, err))
+				}
+			} else {
+				err = tx.Commit()
+			}
+		}()
+
+		return err
+	}
```
zwass: I think this `defer` needs to come before calling the `fn`. Otherwise we miss catching any panics that happen in the function.

```suggestion
		defer func() {
			if p := recover(); p != nil {
				tx.Rollback()
				panic(p)
			} else if err != nil {
				rbErr := tx.Rollback()
				if rbErr != nil && rbErr != sql.ErrTxDone {
					panic(fmt.Sprintf("got err '%s' rolling back after err '%s'", rbErr, err))
				}
			} else {
				err = tx.Commit()
			}
		}()

		return err
	}
	
	err = fn(tx)
```

server/datastore/mysql/mysql.go
```diff
@@ -50,6 +51,40 @@ func (d *Datastore) getTransaction(opts []kolide.OptionalArg) dbfunctions {
...
+			if p := recover(); p != nil {
+				tx.Rollback()
+				panic(p)
+			} else if err != nil {
+				rbErr := tx.Rollback()
+				if rbErr != nil && rbErr != sql.ErrTxDone {
+					panic(fmt.Sprintf("got err '%s' rolling back after err '%s'", rbErr, err))
+				}
+			} else {
+				err = tx.Commit()
```
zwass: `errors.Wrap` here.

server/datastore/mysql/mysql.go
billcobbler: Ah, good call. I think we'll need to split it and only handle the panic in the defer otherwise `err` will only contain results from the d.db.Beginx() call.

billcobbler: Pushed up what I think it should look like by only handling the panic in the defer.

PR #2217, opened 2020-04-01
Remove extra mutex unlock in inmem.AddHostToPack

twexler:
On https://github.com/kolide/fleet/blob/master/server/datastore/inmem/packs.go#L129, there is an extra call to `d.mtx.Unlock()`, since there is a `defer` on https://github.com/kolide/fleet/blob/master/server/datastore/inmem/packs.go#L125, this Unlock actually causes a panic when hitting the `return` statement


zwass:
I am going to close this as the CLA remains unsigned and we are removing the inmem storage layer anyway.

PR #2224, opened 2020-04-13
Sanitize column names for order key in MySQL

zwass:
This adds a SQL injection prevention for a case in which we cannot use
parameters in the query.

It is not clear that this was possible to exploit. If it was possible,
it would have required a valid login to the Fleet server.


zwass:
I think regex is mitigated by the fact that it's in line with a network call to a database. That should dramatically outweigh any perf hit of the regex.

Issue #1439, opened 2017-03-20
Implement Leveled Logging


murphybytes:
https://godoc.org/github.com/go-kit/kit/log/level

Right now we have a course grained logging system that probably incurs a substantial performance overhead on a busy server.  In addition, it's hard to filter out noise to find significant events.  We should implement a leveled logging system where the default is to log only errors and possibly info messages,

marpaia:
In #1629, @gjvc said

> ### What did you expect to see?
>
> the ability to enable / disable logging output as required for debugging or general operation.
>
> ### What did you see instead?
>
> Right now, the fleet server is quite chatty on stdout, and there's no way of hushing it up.
>
> ### implementation suggestion
>
> I would set all current log messages at debug level, so that new ones at info are the only ones visible under normal operations.

Mazon:
We are using fleet in kubernetes and having stdout log to graylog and i agree, having anything other then results from scheduled scans and errors is making it difficult for us to use it.

zwass:
I can see a couple of solutions to this:

1) Consider all "request" logs to be debug logs and then set the default log level to info.

2) Log "request" logs to a separate logger, so that these can be toggled independently of other debug logging (and potentially go to a different output).

I'm inclined towards (1) right now as it would be straightforward in terms of configuration and quick to implement.

crimsonknave:
> We are using fleet in kubernetes and having stdout log to graylog and i agree, having anything other then results from scheduled scans and errors is making it difficult for us to use it.

This is exactly what's happening to me, I'd love to see a fix for this. I'd be perfectly happy to have all the "request" logs be debug. As it is, I didn't notice any difference when I turned debug off (unless removing the env var `KOLIDE_LOGGING_DEBUG` didn't do what I thought it would.)

TransactCharlie:
I've raised a PR here: https://github.com/kolide/fleet/pull/2225 . It's a simple modification to move all the logging middleware (essentially debug for each time a handler is called) to the `debug` level unless there actually was an error.

zwass:
Closing this with #2225. That will reduce log verbosity substantially for default settings.

PR #2155, opened 2019-11-20
Improvements to server logging

thehellajeff:
Currently fleet logging doesn't allow for any amount of auditing, in referencing #2145 fleet should at least log what commands are run by which user

- Added logging for new campaigns
- Added logging for new query creations/modification/deletion
- Added usernames for logs found in labels, options, packs, osquery options, queries and scheduled queries where something is created, modified or deleted

If there are any other ideas for helping with logging please let me know.

PR #2225, opened 2020-04-15
Set all handlers to log at debug level

TransactCharlie:
Kolide fleet logs *a lot* and in particular the handler logs are only questionably valuable.

In k8s for example these logs quickly fill up and can even overwhelm the host they are running on if you have a lot of assets continually talking to `kolide-fleet`

This preserves the log messages generated for each handler but makes them DEBUG level. This should get filtered out by the command line switches. unless `--logging_debug` is passed.

If there *was* an actual error the log lines will be at the `ERROR` level.

There are a few threads expressing a similar desire: https://github.com/kolide/fleet/issues/1439


server/service/logging_appconfig.go
```diff
@@ -14,7 +14,7 @@ func (mw loggingMiddleware) NewAppConfig(ctx context.Context, p kolide.AppConfig
 	)
 
 	defer func(begin time.Time) {
-		_ = mw.logger.Log(
+		_ = mw.loggerForError(err).Log(
```
directionless: The usual way we've done this, is to call level right here. 

`level.Debug(mw.logger).Log(...)`

But I'm going to defer to @zwass on what he'd like

zwass: Typically I'd agree with you @directionless, but in this case it seems that we would have to repeat the `if` statement in `loggerForError` (https://github.com/kolide/fleet/pull/2225/files#diff-4ad2cc4cc27fc85e05f4ba0a04b1d080R22-R25) dozens of times if we wanted to achieve the same effect. What do you think?

TransactCharlie: yeah that was why I wrote that helper receiver on the struct. It would be a real mess of code repeated in every `defer` to decide wether debug or error level should be used and I like the fact that it would upgrade an actual error to an error line

TransactCharlie: maybe the receiver could have a better name though. I suck at naming things.

directionless: Yeah... That's true. This is a fairly elegant way to roll together debug trace level logs with error level logs.

server/service/logging.go
```diff
@@ -15,3 +16,11 @@ type loggingMiddleware struct {
 func NewLoggingService(svc kolide.Service, logger kitlog.Logger) kolide.Service {
 	return loggingMiddleware{Service: svc, logger: logger}
 }
+
+// loggerForError returns a logger with a log level dependant on error
+func (lm *loggingMiddleware) loggerForError(err error) kitlog.Logger {
+	if err != nil {
+		return level.Error(lm.logger)
```
groob: I prefer Info vs Error. 

https://dave.cheney.net/2015/11/05/lets-talk-about-logging

TransactCharlie: well there was an error. I guess it depends how that error is in turn handled but the logger doesn't have any visibility of that. Filtering in something like kibana is a lot nicer when you can say: 'level = error' rather than something more complex like 'err != "nill" - it feels more consistent in a multi tenanted environment with different systems feeding into the same log collector

For my use case I don't much mind -- I just need the mass of useless log lines triggered by every single asset doing anything with kolide-fleet gone.  If it's a blocker Im totally fine with this being Info.

TransactCharlie:
> I think the most important thing we need to do before merging this is find a way to continue outputting the logs from #2155. Many folks are interested in having some sort of audit trail of user actions, and while that work is not complete it does seem to satisfy some needs.
> 
> It's possible that this could be achieved by just dropping the changes on the logging for those methods.
> 
> Please feel free to hop on osquery Slack in #kolide. We can probably get your local build sorted in just a few minutes.

Ok -- so I think I understand what you mean. For clarity:

I've set *every* handler to log here at debug (or error if there was an error). However it would be better for the handlers called out in https://github.com/kolide/fleet/pull/2155 to be always at at least the INFO level?

I think that would be completely fine for my use case. I can probably work on that today.

TransactCharlie:
Latest commit refactors the level logic to make it a lot more obvious what's going on. Also I've set everything I could see that would be an `admin` type action -- or one that you'd want to know about from a sysadmin point of view to INFO. The rest of the handlers log at the DEBUG level to avoid log spam.

TransactCharlie:
@zwass -- ping just in case you didn't get a notification on your requests changes.

server/service/logging.go
TransactCharlie: also I don't see this as *handling* the error in the classic sense. These log events are deffered after the handler call has completed and really just provide observability. The actual error may or may not be handled futher up in the call stack (I have no idea if some other part of middleware) swallows the error or hadles it in some other way.

If I was writing code inline I'd agree with the contents of the link you posted. - i.e something like this it wouldn't make sense to be `log.Error` beacause you really are just continuing -- it's obviously not that big a deal.

```golang
foo, err := doSomething()
if err != nil {
    // Swalling this error because it isn't really a problem
    log.Info("Something something")
}
...
```

I don't have any context of what *happens* to the errors this PR is talking about though and as an admin of the system I'd like to see them differently from more `admin` level info messages.

groob: I do agree with some of what you're saying. It's ultimately up to @zwass to make a choice here.

For context, you don't really care about the logging here, so maybe that's an argument for dropping even errors to debug. They're handled by the http response, which _should_ (not necessarily happening in this code) be also logging the response to the client. 
It's important to care for an admin what got sent to the client ( status + reason) and perhaps alert on those. In the middleware it's more important to the developer.

TransactCharlie:
OK I'm fine with that. Removed the logic to escalate an err => error level.

zwass:
Thank you @TransactCharlie :)

TransactCharlie:
:Woot!: -- thankyou for reviewing.

PR #2229, opened 2020-04-17
Explicitly disable CGO in build

zwass:
Fleet does not use CGO. In some build situations, not explicitly
disabling CGO results in binaries that have linking errors when running
on an Alpine Docker container. This change should fix the build for
those containers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,10 +107,10 @@ endif
 build: fleet fleetctl
 
 fleet: .prefix .pre-build .pre-fleet
-	go build -tags full -i -o build/${OUTPUT} -ldflags ${KIT_VERSION} ./cmd/fleet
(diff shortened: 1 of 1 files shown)


zwass:
I agree, though looks like that was attempted previously (https://github.com/kolide/fleet/blob/master/Makefile#L99) and didn't work. Unfortunately I don't have the bandwidth right now to find the cleanest way to do it so the explicit approach will be used here.

directionless:
I would not expect setting variables in a rule to work like that. But, I assume,  `APP_NAME` does. Skimming the docs, I think that `APP_NAME` is a make variable, and eval is can effect make level things. 

But, `CGO_ENABLED` and `GOGC` are shell level variables, and setting them in a line like that has no effect. (by default, each line of a rule is a _new shell_.

If it's meant to be global, the correct thing is to export it toplevel, like https://github.com/kolide/fleet/blob/master/Makefile#L3

PR #2239, opened 2020-05-25
Update test Docker container versions

zwass:
Previously these test Docker containers used
https://github.com/kolide/docker-osquery, which no longer seems to be
maintained and has very old osquery versions (2.7.0). Migrate to using
containers from the new https://github.com/dactivllc/docker-osquery
project. This brings Fleet to testing with current (4.3.0) osquery on an
expanded set of operating systems.

### DIFF
--- a/tools/osquery/docker-compose.yml
+++ b/tools/osquery/docker-compose.yml
@@ -3,7 +3,7 @@ version: '2'
 
 services:
   ubuntu14-osquery:
-    image: "kolide/osquery:${KOLIDE_OSQUERY_VERSION}"
+    image: "dactiv/osquery:4.3.0-ubuntu14.04"
     volumes:
       - ./kolide.crt:/etc/osquery/kolide.crt
       - ./example_osquery.flags:/etc/osquery/osquery.flags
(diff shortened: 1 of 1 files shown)


tools/osquery/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: '2'
 
 services:
   ubuntu14-osquery:
-    image: "kolide/osquery:${KOLIDE_OSQUERY_VERSION}"
+    image: "dactiv/osquery:4.3.0-ubuntu14.04"
```
directionless, line 6: Given that this is in the `kolide` org, it may make more sense to fix `kolide/osquery`? Thoughts

zwass: If you would like to maintain `kolide/osquery`, that's fine with me. The important bit IMO is that we are developing with new versions of osquery.

zwass: @directionless do you think Kolide can take on maintaining `kolide/osquery`?

directionless: Maybe. Want to help? I can give you permission on both the git side, and the docker side...

PR #2244, opened 2020-06-17
Refactor `fleetctl get` subcommands so that they all have json and yaml output options

grantseltzer:
In this proposed patch all subcommands now have a new `--json` flag to allow for json output. They will default to using an ascii table if that was previously the default. 

Signed-off-by: grantseltzer <grantseltzer@gmail.com>


grantseltzer:
@zwass Absolutely, good catch. They should all have consistent defaults. Will push new commit in the morning.

grantseltzer:
@zwass Went ahead and made that change 👍

grantseltzer:
Thanks for the merge!

PR #2258, opened 2020-07-07
Clean up unused DB tables, migrations, and code

zwass:
This PR removes unused types, code, DB tables, and associated migrations that are unused since Fleet 2.0.

An existing migration was refactored, and should remain compatible with both existing and new Fleet installations.


server/datastore/mysql/migrations/tables/20171116163618_CreateTableOsqueryOptions.go
```diff
@@ -15,7 +25,137 @@ func Up_20171116163618(tx *sql.Tx) error {
...
+		if err != nil {
+			return errors.Wrap(err, "insert options")
+		}
+	}
+
+	return nil
+}
+
+func migrateOptions(tx *sql.Tx) error {
+	type configForExport struct {
```
zwass, line 63: This is just moved from https://github.com/kolide/fleet/pull/2258/files#diff-b49dfacbb915d42e49cab4e6b35cad44L26

server/kolide/osquery.go
```diff
@@ -78,27 +78,3 @@ type PermissivePackContent struct {
...
-// Decorators is the format of the decorator configuration in an osquery config.
-type Decorators struct {
-	Load     []string            `json:"load,omitempty"`
-	Always   []string            `json:"always,omitempty"`
-	Interval map[string][]string `json:"interval,omitempty"`
-}
-
-// OsqueryConfig is a struct that can be serialized into a valid osquery config
-// using Go's JSON tooling.
-type OsqueryConfig struct {
```
directionless, line 91: Feels weird to lose this. Though it's probably better in the go sdk anyhow.

zwass: It's actually been unused since Fleet 2.0. It's up to the user to put in a valid config (via yaml) and we don't have to do anything to it besides serialize to JSON.

PR #2282, opened 2020-08-07
add stdout and kinesis logger plugins and add sts assume role to firehose

billcobbler:
- new stdout osquery logger plugin
- new kinesis osquery logger plugin
- add sts assume role support to firehose osquery logger plugin

### DIFF
--- a/docs/infrastructure/configuring-the-fleet-binary.md
+++ b/docs/infrastructure/configuring-the-fleet-binary.md
@@ -539,7 +539,7 @@ Setting this to a higher value can reduce baseline load on the Fleet server in l
 
 Which log output plugin should be used for osquery status logs received from clients.
 
-Options are `filesystem`, `firehose`, and 'pubsub'.
+Options are `filesystem`, `firehose`, `kinesis`, `pubsub`, and `stdout`.
 
 - Default value: `filesystem`
 - Environment variable: `KOLIDE_OSQUERY_STATUS_LOG_PLUGIN`
@@ -554,7 +554,7 @@ Options are `filesystem`, `firehose`, and 'pubsub'.
 
 Which log output plugin should be used for osquery result logs received from clients.
 
-Options are `filesystem`, `firehose`, and 'pubsub'.
+Options are `filesystem`, `firehose`, `kinesis`, `pubsub`, and `stdout`.
 
 - Default value: `filesystem`
 - Environment variable: `KOLIDE_OSQUERY_RESULT_LOG_PLUGIN`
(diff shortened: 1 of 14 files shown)


zwass:
Thanks for submitting this PR! I'll give it a review early next week.

gavinelder:
Drive-by thought when looking at this PR and the addition of `stdout` I wonder how easy it would be to support multiple destinations at the same time. 

I see a use case where someone may want `stdout` for debug but also want `kinesis` to also be running for their logging pipeline vs them being mutually exclusive.

docs/infrastructure/configuring-the-fleet-binary.md
```diff
@@ -765,6 +780,12 @@ Name of the Firehose stream to write osquery status logs received from clients.
 		status_stream: osquery_status
 	```
 
+The IAM role used to send to Firehose must allow the following permissions on
+the stream listed:
+
+* `firehose:DescribeDeliveryStream`
```
zwass, line 786: Nice, this is helpful documentation.

docs/infrastructure/working-with-osquery-logs.md
```diff
@@ -40,3 +55,14 @@ Note that Firehose logging has limits [discussed in the documentation](https://d
...
 Note that messages over 10MB will be dropped, with a notification sent to the fleet logs, as these can never be processed by PubSub.
+
+### Stdout
+
+- Plugin name: `stdout`
+- Flag namespace: [stdout](../infrastructure/configuring-the-fleet-binary.md#stdout)
+
+With the stdout plugin, osquery result and/or status logs are written to stdout
+on the Fleet server. This is typically used for debugging or with a log
+log forwarding setup that will capture and forward stdout logs into a logging
```
zwass: extraneous "log"

server/logging/firehose.go
```diff
@@ -49,6 +50,17 @@ func NewFirehoseLogWriter(region, id, secret, stream string, logger log.Logger)
 	if err != nil {
 		return nil, errors.Wrap(err, "create Firehose client")
 	}
+
+	if stsAssumeRoleArn != "" {
```
zwass, line 54: Can we refactor this a bit so that there is only a single call to `session.NewSession` and above that is something like
```
if stsAssumeRoleArn != "" {
    creds := stscreds.NewCredentials(sess, stsAssumeRoleArn)
    conf.Credentials = creds
}
```

server/logging/kinesis.go
```diff
@@ -0,0 +1,204 @@
...
+	if id != "" && secret != "" {
+		conf.Credentials = credentials.NewStaticCredentials(id, secret, "")
+	}
+
+	sess, err := session.NewSession(conf)
+	if err != nil {
+		return nil, errors.Wrap(err, "create Kinesis client")
+	}
+
+	if stsAssumeRoleArn != "" {
```
zwass, line 56: Same refactor here please

server/logging/kinesis.go
```diff
@@ -0,0 +1,204 @@
...
+	out, err := k.client.DescribeStream(
+		&kinesis.DescribeStreamInput{
+			StreamName: &k.stream,
+		},
+	)
+	if err != nil {
+		return errors.Wrapf(err, "describe stream %s", k.stream)
+	}
+
+	if (*(*out.StreamDescription).StreamStatus) != kinesis.StreamStatusActive {
```
zwass, line 94: Do we need nil checks here? It's not immediately clear to me that this are guaranteed to be non-nil.

server/logging/firehose.go
nyanshak: As far as I can tell from reading docs about aws-sdk-go, I don't think this is possible, but I'm not super familiar with it.

AFAICT you need to have an existing AWS session to get sts credentials.

server/logging/kinesis.go
nyanshak: AWS's API says that it should return StreamStatus with this type of request. It's marked as 'required' in the Go code. If we don't get a non-nil error, I don't think it's possible to have these be nil.

We could add them, but I don't think it's needed. (aside: This follows the same pattern as seen in the existing Firehose logger.)

nyanshak:
```
Drive-by thought when looking at this PR and the addition of stdout I wonder how easy it would be to support multiple destinations at the same time.

I see a use case where someone may want stdout for debug but also want kinesis to also be running for their logging pipeline vs them being mutually exclusive.
```

@gavinelder - good suggestion. Trying to keep this PR relatively focused, but maybe that could be a separate issue if you're interested.

docs/infrastructure/working-with-osquery-logs.md
nyanshak: updated

gavinelder:
> ```
> Drive-by thought when looking at this PR and the addition of stdout I wonder how easy it would be to support multiple destinations at the same time.
> 
> I see a use case where someone may want stdout for debugging but also want kinesis to also be running for their logging pipeline vs them being mutually exclusive.
> ```
> 
> @gavinelder - good suggestion. Trying to keep this PR relatively focused, but maybe that could be a separate issue if you're interested.

I will cut an issue / FR for this

server/logging/firehose.go
zwass: Apologies for the delay. What I was referring to was using a strategy like on lines 45-57.

Something like this on line 48

```
if stsAssumeRoleArn != "" {
    creds := stscreds.NewCredentials(sess, stsAssumeRoleArn)
    conf.Credentials = creds
}
```

Then we have only one place where `sess, err = session.NewSession(conf)` appears.

zwass: I missed that creating the credentials requires an already active session. This looks good.

Issue #2180, opened 2020-01-13
Feature Request: Add Log Rotation Compress Option


hackdefendr:
1. This is a simple feature request for adding --filesystem_enable_log_compress to compress logs that are rotated.  Much in the same way logrotate does to save disk space. As of now, the log rotation seems to occur when the logs reach 5mb in size.  Compressing these rotated files reduces them to well below 1mb in size.

2. Possibly adding to this, make an option for --filesystem_enable_log_history to tell fleet how many rotated logs to keep before purging files.

Does this make sense?

zwass:
Request 1 was implemented by #2292. If you are still interested in Request 2, please open a new issue.

PR #2292, opened 2020-08-26
Add compression option for filesystem logs when they're rotated

jalseth:
This adds support for compressing the rotated log files if the `filesystem` logger plugin is set.

This should address the first portion of #2180.


zwass (approved):
Nice, thank you!

Issue #2301, opened 2020-09-17
Feature request: `fleetctl get hosts` should show osquery version


jalseth:
Currently `fleetctl get hosts` only shows the host UUID, hostname, platform, and online status in the table. Another useful item would be to show the version of osquery that the host is running. 

This can currently be worked around by using the `--json` flag and piping that to `jq` to parse the information, but it would be nice to not have to do this.

### What version of `fleet` are you using (`fleet version --full`)?

3.1.0

### What operating system are you using?

macOS

### What did you do?

`fleetctl get h`

### What did you expect to see?

Host information, including osquery version.

### What did you see instead?

UUID, hostname, platform, and online status only.

zwass:
This would be pretty easy to add. At some point I imagine we will want to make the outputs of that command configurable, but for now I think it would be fine to just add the new column in the code.

PR #2304, opened 2020-09-18
Add osquery_version field to `fleetctl get hosts`

jalseth:
This change makes the `fleetctl get hosts` command include the osquery version in the output. It also fixes another string(int) compatibility issue with Go 1.15. 

Fixes https://github.com/kolide/fleet/issues/2301


zwass (changes requested):
@jalseth can you please split these changes into two separate PRs?

jalseth:
@zwass I updated the commits here

zwass (approved):
Thanks!

PR #2305, opened 2020-09-18
Add -L option to curl

res0nat0r:
The example in the README won't follow the github redirect without `-L`, so the example as is will download an html file instead of the actual `fleet.zip`. This fixes that.


zwass (approved):
Thank you!

Issue #2287, opened 2020-08-19
make build broken on 3.0.0 and higher


jalseth:
The builds fail on 3.0.0 and higher. This appears to be due to the addition to the `CGO_ENABLED=0` environment variable added to all of the build commands. If I remove this, the build is successful.

### What version of `fleet` are you using (`fleet version --full`)?

N/A

### What operating system are you using?

Tested on macOS Catalina 10.15.6 with Golang 1.15.0 and Ubuntu 20.04 with Golang 1.14.2

### What did you do?

```
$ make deps
$ make generate
$ make build
```

### What did you expect to see?

Successful build.

### What did you see instead?

```
$ make build
mkdir -p build/linux
mkdir -p build/darwin
CGO_ENABLED=0 go build -tags full -i -o build/fleet -ldflags " -X github.com/kolide/kit/version.appName=fleet -X github.com/kolide/kit/version.version=3.0.0 -X github.com/kolide/kit/version.branch=HEAD -X github.com/kolide/kit/version.revision=0058d459022df132df36556c40c342958639501c -X github.com/kolide/kit/version.buildDate=2020-08-19T19:23:52Z -X github.com/kolide/kit/version.buildUser=james -X github.com/kolide/kit/version.goVersion=go1.15" ./cmd/fleet
go build net: copying /Users/james/Library/Caches/go-build/0b/0bea6c310e5c2fb4021cd6e334c52f7b692fea5ff47dcb3af72b710843a8afc1-d: open /opt/local/lib/go/pkg/darwin_amd64/net.a: permission denied
go build os/user: copying /Users/james/Library/Caches/go-build/1f/1fc0c83ac80ef7f361f8f91ae0828b4b8a43f73a00b4dfb253c529074e30a990-d: open /opt/local/lib/go/pkg/darwin_amd64/os/user.a: permission denied
make: *** [fleet] Error 1
```

zwass:
Is this with Go 1.15? I've not seen such issues before. Maybe because I have `GOPATH` set?

jalseth:
As noted in the issue, I tried both Go 1.15 and 1.14 on macOS and Ubuntu respectively. 

This is from a fresh Ubuntu 18.04 VM. I have skipped the output on the successful steps.

```
$ sudo apt update && sudo apt upgrade -y && sudo apt install -y git npm
$ sudo npm install -g yarn
$ curl -L -o go.tar.gz https://golang.org/dl/go1.14.7.linux-amd64.tar.gz
[...]

PR #2307, opened 2020-09-25
Remove -i flag from Go builds in the Makefile

jalseth:
The `-i` (install) flag is not necessary and caused issues with building in some scenarios.

Fixes https://github.com/kolide/fleet/issues/2287

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,10 +107,10 @@ endif
 build: fleet fleetctl
 
 fleet: .prefix .pre-build .pre-fleet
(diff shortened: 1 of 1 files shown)


zwass (approved):
I took a look through the history to find where that flag was added (https://github.com/kolide/fleet/pull/197/files#diff-b67911656ef5d18c4ae36cb6741b7965R65) and can't see any indication that it was added for a specific reason. 

I successfully tested a local build on macOS as well as a cross-compile for the Docker container. Both seemed to generate working binaries.

PR #2309, opened 2020-09-26
Fix string(int) in Kinesis logger for Go 1.15 compatibility

jalseth:
Go 1.15 does not allow string(int) type casting.


zwass (approved):
Thank you!

Issue #2233, opened 2020-04-29
Delete Host Confirmation Prompt Does not Show Hostname


jalseth:
### What version of `fleet` are you using (`fleet version --full`)?

2.6.0

### What operating system are you using?

N/A

### What did you do?

Delete an offline host from the Fleet UI

### What did you expect to see?

The prompt includes the host name

### What did you see instead?

The prompt just warns that you are deleting a host, not which one you are deleting

### Additional Notes

This isn't as much of an issues when the hosts are in grid view, but in list view it is easier to accidentally click on the wrong trash can icon.

PR #2310, opened 2020-09-27
Add hostname to delete host confirmation modal

jalseth:
Fixes https://github.com/kolide/fleet/issues/2233

### DIFF
--- a/frontend/pages/hosts/ManageHostsPage/ManageHostsPage.jsx
+++ b/frontend/pages/hosts/ManageHostsPage/ManageHostsPage.jsx
@@ -341,7 +341,7 @@ export class ManageHostsPage extends PureComponent {
         onExit={toggleDeleteHostModal(null)}
         className={`${baseClass}__modal`}
       >
-        <p>This action will delete the host <strong>${selectedHost.hostname}</strong> from your Fleet instance.</p>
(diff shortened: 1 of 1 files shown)


frontend/pages/hosts/ManageHostsPage/ManageHostsPage.jsx
```diff
@@ -341,7 +341,7 @@ export class ManageHostsPage extends PureComponent {
         onExit={toggleDeleteHostModal(null)}
         className={`${baseClass}__modal`}
       >
-        <p>This action will delete the host from your Fleet instance.</p>
+        <p>This action will delete the host <strong>${selectedHost.hostname}</strong> from your Fleet instance.</p>
```
zwass: Is the `$` intentional?

```suggestion
        <p>This action will delete the host <strong>{selectedHost.hostname}</strong> from your Fleet instance.</p>
```

zwass (changes requested):

frontend/pages/hosts/ManageHostsPage/ManageHostsPage.jsx
jalseth: Apologies, I don't have much experience with frontend languages. It was my understanding that it was needed when inserting the variable to a string, but now that I'm looking at this again this isn't actually in a string block, so it isn't needed.

zwass (approved):

PR #2311, opened 2020-10-01
Add redis use_tls cfg

### LINKED ISSUE #2247: Kolide Fleet to Redis SSL/TLS connection option
Is there an SSL/TLS option for connecting Kolide Fleet to a Redis instance? 
I am trying to configure Kolide Fleet to use an Azure Cache for Redis instance that has only an SSL connection option only. 

fleet - version 2.6.0
  branch: 	master
  revision: 	d3849d8fe1ea841d34a19e5cc06bfe62fc8dfd2d
  build date: 	2020-03-24T17:27:31Z
  build user: 	zwass
  go version: 	go1.13.4

Red Hat Enterprise Linux 7

### What did you do?

Configure fleet to use an Azure Cache for Redis instance that has only an SSL connection option only. 

Live query result

### What did you see instead?

Live Queue does not work, seeing the below error instead.

Jun 23 14:15:44 xxxx.xxx.xxx fleet[109099]: {"component":"service","err":"reading from redis: read tcp xx.xx.xxx.xxx:36706-\u003exx.xx.xxx.xxx:6380: read: connection reset by peer","method":"NewDistributedQueryCampaign","numHosts":0,"sql":"SELECT * FROM osquery_info","took":"1m45.777923261s","ts":"2020-06-23T21:15:44.110580311Z","user":"xxxxxxx"}

kilianw:
Adding config parameter 'redis.use_tls' to enable tls communications with redis e.g. AWS ElastiCache

Closes #2247

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -37,7 +37,7 @@ type RedisConfig struct {
 	Address  string
 	Password string
 	Database int
-	UseTLS   bool
+	UseTLS   bool `yaml:"use_tls"`
 }
 
 const (
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -37,6 +37,7 @@ type RedisConfig struct {
 	Address  string
 	Password string
 	Database int
+	UseTLS   bool
 }
 
 const (
@@ -182,6 +183,7 @@ func (man Manager) addConfigs() {
 		"Redis server password (prefer env variable for security)")
 	man.addConfigInt("redis.database", 0,
 		"Redis server database number")
+	man.addConfigBool("redis.use_tls", false, "Redis server enable TLS")
 
 	// Server
 	man.addConfigString("server.address", "0.0.0.0:8080",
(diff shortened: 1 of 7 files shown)


zwass:
Looks like this should fix #2247?

server/config/config_test.go
```diff
@@ -47,7 +47,12 @@ func TestConfigRoundtrip(t *testing.T) {
 			case int:
 				key_v.SetInt(int64(conf_index*100 + key_index))
 			case bool:
-				key_v.SetBool(true)
+				switch conf_v.Type().Field(key_index).Name {
+				case "UseTLS":
```
zwass: I believe that this test failure actually caught an issue with how the configuration is handled. I am digging into this.

server/config/config.go
```diff
@@ -37,6 +37,7 @@ type RedisConfig struct {
 	Address  string
 	Password string
 	Database int
+	UseTLS   bool `yaml:"use_tls"`
```
zwass, line 40: I added the `yaml` tag here which fixes the config file parsing (and related test).

zwass (approved):
Thank you!

PR #2319, opened 2020-10-07
docker-compose QoL improvements

### LINKED ISSUE #2127: Can't start osqueryd test instances via docker-compose on Linux

d453ca3 (master)

Linux

```
uname -a
Linux <hostname> 5.3.7-arch1-2-ARCH #1 SMP PREEMPT @1572002934 x86_64 GNU/Linux
```

### What did you do?

Following `tools/osquery/README.md`:

```
export KOLIDE_OSQUERY_VERSION="latest"
export ENROLL_SECRET="<secret>"
docker-compose up
```

4 test hosts connecting to fleet locally

### What did you see instead?

```
centos7-osquery_1   | W1101 02:15:39.805672     8 tls_enroll.cpp:75] Failed enrollment request to https://host.docker.internal:8080/api/v1/osquery/enroll (Request error: Host not found (authoritative)) retrying...
```

It looks like previously, this file used `extra_hosts` block in compose file to add an extra host called `dockerhost` set to the env var `$LOCALHOST`, then used `dockerhost` in the `tls_hostname` in the flag file.

In #2019, this was changed to use `host.docker.internal` instead, with `tls_hostname` set to `host.docker.internal`, to enable starting osquery instances with less configuration.

However, Docker on Linux *does not* add the `host.docker.internal` DNS name, so this prevents starting the osqueryd test containers from connecting to fleet on Linux machines.

nyanshak:
* Use YAML anchors to avoid repeating config blocks
* Use docker volumes to persist data for mysql
* Allow setting `FLEET_SERVER` (fixes #2127) when using the docker-compose file to spin up multiple osquery clients

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   mysql:
     image: mysql:5.7
     volumes:
-      - .:/tmp
+      - mysql-persistent-volume:/tmp
     command: mysqld --datadir=/tmp/mysqldata --slow_query_log=1 --log_output=TABLE --log-queries-not-using-indexes --event-scheduler=ON
     environment: &mysql-default-environment
       MYSQL_ROOT_PASSWORD: toor
@@ -44,3 +44,6 @@ services:
       - /var/run:/var/run:rw
       - /sys:/sys:ro
       - /var/lib/docker/:/var/lib/docker:ro
+
+volumes:
+  mysql-persistent-volume:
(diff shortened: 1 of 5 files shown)


nyanshak:
I was also going to ask about updating the test images (to use newer operating systems / osquery versions), but I see there's already a convo about that here: https://github.com/kolide/fleet/pull/2239.

zwass (approved):
Nice, thank you for this!

I tested in my macOS dev environment and everything works as expected.

PR #2320, opened 2020-10-07
#2283: Consistent use of constants in fleetctl

### LINKED ISSUE #2283: Refactor fleetctl get and apply to use constants for spec kinds
Use constants for the kinds in 
https://github.com/kolide/fleet/blob/master/cmd/fleetctl/get.go#L67, etc.
https://github.com/kolide/fleet/blob/master/cmd/fleetctl/apply.go#L53, etc.
to ensure that we avoid bugs in the future such as https://github.com/kolide/fleet/pull/2281.

nyanshak:
Refactor fleetctl get & apply to use constants for spec kinds, fix #2283


zwass (approved):
Nice, thank you for tackling this!

Issue #2240, opened 2020-06-04
fleetctl feature - add ability to import packs as disabled


defensivedepth:
Would like to have the ability to import packs disabled by default instead of enabled.

PR #2325, opened 2020-10-09
fleetctl: allow disabling packs via apply

nyanshak:
* Fixes issue #2240

### DIFF
--- a/cmd/fleetctl/get.go
+++ b/cmd/fleetctl/get.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/ghodss/yaml"
 	"github.com/kolide/fleet/server/kolide"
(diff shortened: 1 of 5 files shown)


nyanshak:
```
fleetctl get packs
+--------------------+----------+-------------+----------+
|        NAME        | PLATFORM | DESCRIPTION | DISABLED |
+--------------------+----------+-------------+----------+
| example-pack       |          |             | true     |
+--------------------+----------+-------------+----------+

fleetctl get packs --yaml
---
kind: pack
spec:
  disabled: false
  id: 20
  name: example-pack
  queries:
    - description: ""
      interval: 30
      name: example-query
      query: example-query
  targets:
    labels:
      - All Hosts

fleetctl get pack osquery-monitoring --with-queries
# same output as previous command with the queries:
...
---
apiVersion: v1
kind: query
spec:
  description: example query
  name: example-query
  query: SELECT build_platform FROM osquery_info;
```

zwass (changes requested):
Nice, definitely a sensible change. Before we merge this, can you update the test to cover the disabled case? It could probably be as simple as adding a disabled pack in https://github.com/kolide/fleet/blob/master/server/datastore/datastore_packs_test.go#L153.

zwass (approved):
Thank you!

Issue #2146, opened 2019-11-12
Remove soft-deletion pattern from Fleet


zwass:
As part of the transition from Fleet 1.x to Fleet 2.x we mostly moved off of the soft-deletion pattern. Most of the code was refactored to cease using this, and soft-deleted entities were purged in a migration (https://github.com/kolide/fleet/pull/1959/files).

In this process, we missed a couple of places where soft deletion is still used (https://github.com/kolide/fleet/blob/06832697d0e6ed6b2ca0220ef5434791db7b0a27/server/datastore/mysql/scheduled_queries.go#L115, https://github.com/kolide/fleet/blob/c8229cc0d63e07e60517d8391637209b4e965387/server/datastore/mysql/invites.go#L129). These are probably best refactored by using regular deletion in the `deleteEntity` function, followed by a migration removing any lingering soft deleted entities of these types.

Following this, removing the `deleted` and `deleted_at` columns in the database tables will complete the migration from this pattern, and eliminate any suggestion that this pattern should be used in future development.

PR #2327, opened 2020-10-11
Remove soft-deletion pattern

nyanshak:
* Perform migration to delete any entries with `deleted` set, and
subsequently drop columns `deleted` and `deleted_at`.
* Remove `deleted` and `deleted_at` references.


nyanshak:
Addresses #2146

nyanshak:
@zwass I took a stab at solving #2146 here. It's still a bit of a work in progress, but I could use some help with the migrations since I haven't worked with that before.

I'm following the instructions in `docs/development/database-migrations.md`, but having a bit of trouble.

When I run `./build/fleet prepare db`, my migration file is called (added debug prints in init), but the `Up` function doesn't seem to ever be called.

Could you take a look and see if I'm doing something wrong?

server/datastore/mysql/migrations/data/20171212182459_DeleteSoftDeletedPacks.go
```diff
@@ -1,23 +0,0 @@
-package data
```
zwass, line 1: Deleting migrations that have been published in a release can lead to inconsistencies. We need to leave these migrations and assume they have been run prior to the new migration running.

server/datastore/mysql/migrations/data/20181119180000_DeleteSoftDeletedEntities.go
```diff
@@ -1,54 +0,0 @@
-package data
```
zwass, line 1: Deleting migrations that have been published in a release can lead to inconsistencies. We need to leave these migrations and assume they have been run prior to the new migration running.

server/datastore/mysql/migrations/tables/20201011162341_CleanupSoftDeletedColumns.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"fmt"
+
+	"database/sql"
+	"github.com/kolide/goose"
+)
+
+func init() {
+	goose.AddMigration(Up20201011162341, Down20201011162341)
```
zwass: This needs to be `MigrationClient.AddMigration`. That will solve your issue with the migration not running. Do we need to update the docs about this?

server/datastore/mysql/migrations/tables/20201011162341_CleanupSoftDeletedColumns.go
```diff
@@ -0,0 +1,95 @@
...
+	deleteStmt := fmt.Sprintf("DELETE FROM `%s` WHERE deleted;", dbTable)
+
+	_, err := tx.Exec(deleteStmt)
+	if err != nil {
+		return err
+	}
+
+	for _, column := range []string{"deleted", "deleted_at"} {
+		alterStmt := fmt.Sprint(
+			"ALTER TABLE `%s` DROP COLUMN `%s`;", dbTable, column)
```
zwass, line 36: I want to point out a couple of concerns about running migrations like this, related to issues you and I have discussed recently:

1) Note that any DDL statement in MySQL (like an `ALTER`) will _implicitly_ end the current transaction. This is the source of the issues of the database ending up in an inconsistent state during migrations. Additionally, MySQL does not support an `IF EXISTS` clause for these statements. It may make sense to try to detect the specific error that could be encountered if the column does not exist and ignore it. In general, I'd like to look towards employing such strategies for DDL migrations.

2) If there are any Fleet servers of the prior version running after these `ALTER`s have executed, they will encounter many errors as they query for the nonexistent columns. I've considered the possibility of using a strategy in which we stop using columns in one release and actually drop them in the next. That would partially mitigate the problem, but additional work would have to be done to explicitly `SELECT` the desired columns so that mapping errors do not occur after new columns are added.

Both of the above problems are mitigated by our new documentation in https://github.com/kolide/fleet/pull/2279/files, but I am also considering what it would take to be able to improve availability during upgrades.

What do you think?

server/datastore/mysql/migrations/tables/20201011162341_CleanupSoftDeletedColumns.go
```diff
@@ -0,0 +1,95 @@
...
+		err := cleanupSoftDeleteFields(tx, table)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func Down20201011162341(tx *sql.Tx) error {
```
zwass, line 107: Fleet doesn't actually have a mechanism for running down-migrations, and most migrations don't actually implement the down portion. Feel free to skip it next time. Do we need to document this?

zwass (changes requested):
**Thank you** for tackling this!

server/datastore/mysql/migrations/tables/20201011162341_CleanupSoftDeletedColumns.go
nyanshak: Ah, yeah, I guess I didn't think about migrating to old versions too much. However, it's interesting that you say most migrations don't implement this. I was looking at the existing migrations and it actually looks like they usually do (or at least the ones I've checked).

Like, if the Up migration adds a column, the Down one will drop the column. Or if Up Creates a table, then Down will drop the table.

I think it would be good to document this either way, but also consider supporting running down-migrations.

In any case - I imagine that's probably out of scope for this PR. I'd just as soon keep the Down migration code in unless we're going to be dropping all of the Down migration functions.

server/datastore/mysql/migrations/data/20171212182459_DeleteSoftDeletedPacks.go
nyanshak: Ah, this really confuses me, based on how I read the the `database-migrations.md` doc.

The section about 'adding/updating tables' says that:

> Once committed to the Fleet repo, table migrations should be considered immutable. Any changes to an existing table should take place in a new migration executing ALTERs.

However, the section on **data** migrations says:

> Data migrations can be mutable. If tables are altered in a way that would render a data migration invalid (columns changed/removed), data migrations should be updated to comply with the new schema.

So, to me, if the `deleted` / `deleted_at` columns no longer exist, that means the migration in the two packs are no longer valid and need to be updated. But it's less clear what that means as far as "updating to comply with the new schema" means.

I'm still not super clear on what I should be doing here instead however.

server/datastore/mysql/migrations/tables/20201011162341_CleanupSoftDeletedColumns.go
nyanshak: The docs say to run `goose create AddColumnFooToUsers`, then 'edit it', but is less clear on what needs to be done.

I see two things that **must** be edited (as far as I can tell) each time you run the `goose create` command:

1. delete the import line for goose:  `github.com/pressly/goose` 
2. change `goose.AddMigration(...)` to `MigrationClient.AddMigration(...)

---

I can add a note to the docs to fix both of these if you'd like.

It would also be good to have the `goose create` command generate this file as close to 'correctly' as possible, so it might be worth updating the goose library to do this.

Also while I was looking into this, I was a bit confused on what the kolide goose fork did, but it seemed like the fork hasn't been updated since 2018, while the pressly source is still actively maintained. Long-term - should we consider moving back to pressly upstream?

server/datastore/mysql/migrations/tables/20201011162341_CleanupSoftDeletedColumns.go
nyanshak: (1) I'll investigate this and see if I can detect the specific case to ignore.

(2) I think this is technically covered by the recommendation to take fleet offline during migrations, but it's always nice to make migrations less painful / maintain higher availability.

I assume when you say 'stop using the columns' in one release, that would mean:
* Do a migration in the first release to drop all rows where `deleted` is true
* Stop using `deleted` fields in select / update statements in the first release, including removing Deleted fields from all the structs.
* In the **next** release, do the ALTER to drop the column.

I think this is a bit better, as it should allow the FIRST release version of fleet to continue operating without errors after the ALTER.

Let's say the releases are as follows:
* x.y.0 - last release before any changes
* x.y.1 - release that stops using delete columns
* x.y.2 - release that alters tables

If someone skips over x.y.1 release directly to x.y.2, they'll still hit this case, but I think that's unavoidable. I'd probably include that in the upgrading guide.

---

I'll look at (1) this weekend, but doing (2) will involve a lot more rework. If you think it would be worthwhile, I can rework this PR into the 'part 1' migration release PR (stop using soft-deleted columns), and make a separate PR later.

nyanshak: Pushing a commit for (1), but it generally looks like:

```
_, err := tx.Exec(stmt)
if err != nil {
    mysqlErr, ok := err.(*mysql.MySQLError)
    if !ok {
        return err
    }

    // depending on whether this is alter or delete, compare against diff constant value
    if mysqlErr.Number != someConstant {
        return err
    }
    fmt.Println("some nice message about skipping because column didn't exist")
}
```

nyanshak: In the new commit, the error messages will look like:
```
Skipped deleting 'soft-deleted' entries from '<table>' because the 'deleted' column does not exist.
Skipped dropping column 'deleted' on table '<table>' because column does not exist.
```

zwass: Yeah I think you have everything pretty much spot-on there.

I think going ahead with (1) and addressing (2) in some later realease (if we decide to) makes sense.

server/datastore/mysql/migrations/tables/20201011162341_CleanupSoftDeletedColumns.go
zwass: You are right that for some time we were writing down-migrations. Eventually we stopped when we realized that we had gone years without needing them or even implementing a way to run them.

I think it's totally fine to include down-migrations in the code if only as documentation. Maybe someday we will remove them.

One of the big problems with actually running down-migrations is that the migrations tend to be lossy and running down-migrations is likely not to return things to the same state. In the unlikely event that the version needs to be downgraded I think it's best for users to restore from a DB backup or manually determine the appropriate commands to run to return the schema to a consistent state.

server/datastore/mysql/migrations/data/20171212182459_DeleteSoftDeletedPacks.go
zwass: Ah, no, you are correct. I had missed that these are data migrations that are no longer valid after the schema changes. Thank you for pushing back.

zwass (changes requested):
It seems there are a couple of test failures that need to be updated.

In regards to your comment about Goose (Github no longer seems to allow me to reply to that comment), please do update the documentation with what you found.

We had initially forked Goose for at least one reason: We needed to be able to make the separate migration "clients" for tables and data, and needed to refactor Goose from using a single, global, migration client. I cannot remember if there were other motivations for the fork.

I did look recently at the activity on Goose and did not get the impression that the core had changed in a way that would allow us to achieve the goal of separate clients. That said, I do think it could be worthwhile to try to rebase the fork onto the new changes.

In regards to making the command generate a migration closer to what we need, that definitely seems worthwhile.

nyanshak:
One of the test failures is when deleting hosts. There is an existing test for `testIdempotentDeleteHost` that creates a host and tries to delete it twice. In the 'soft-deleted' pattern, this was fine (just set `deleted` to true each time). There are lots of places that call `deleteEntity` but `DeleteHost` was the only one with a test expecting this to be idempotent.

I'm not sure what the reasoning is behind that, but I'm going to put up a commit that removes this test, as deleting is not actually idempotent any more. Give me a shout if this doesn't work.

zwass:
I am totally on board with removing tests that no longer make sense. Thank you. I will give this some more review and testing when I have a chance.

zwass (approved):
I just pushed a commit to remove the debug prints. Thank you for sorting this out!

PR #2331, opened 2020-10-13
updated faq with bad record MAC fix suggestion

daniel-infosec:
Updated the FAQ with a fix for the 'bad record MAC' problem that has cropped up in github issues and in the #kolide channel in slack.


zwass (approved):
Thank you!

PR #2336, opened 2020-10-21
Fix form field focus on setup pages

### LINKED ISSUE #936: Focus on first form field during setup step transitions
I think we used to have this but when I go through setup and transition to the next screen (either by clicking the next button or by pressing enter) the first form of the next screen should be auto-focused so I don't have to use my mouse.

noahtalerman:
- Using componentDidUpdate() to check for currentPage change in setup registration form. Initially tried adding `autofocus` prop to the first `<InputFieldWithIcon />` on each page. As seen in AdminDetails page. Didn't work. I believe React only pays attention to `autofocus` when the <input> is re-rendered.

- Calling focus() on page's first input  when currentPage changes and is true. Using refs callback

- Delaying focus by 300ms using setTimeout because the `.user-registration__field-wrapper` has a transition duration of 300ms. Setting the inputs focus immediately creates a snapping movement and ruins the smooth transition.

Note: If using setTimeout isn't ideal, then another potential solution is to create an event to listen for the transition's completion.

Fixes #936

### DIFF
--- a/frontend/components/forms/RegistrationForm/AdminDetails/AdminDetails.jsx
+++ b/frontend/components/forms/RegistrationForm/AdminDetails/AdminDetails.jsx
@@ -25,6 +25,9 @@ class AdminDetails extends Component {
 
   componentDidUpdate(prevProps) {
     if (this.props.currentPage && this.props.currentPage !== prevProps.currentPage) {
+      // Component has a transition duration of 300ms set in 
+      // RegistrationForm/_styles.scss. We need to wait 300ms before 
+      // calling .focus() to preserve smooth transition.
       setTimeout(() => {
         this.firstInput.input.focus();
       }, 300);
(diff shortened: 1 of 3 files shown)


zwass (changes requested):
Clever! Thank you for documenting the reason for the `setTimeout` usage. I was definitely wondering when I saw the diff before I read the PR description. Can you actually put a quick note into the code where that is used so that it's right there for folks to understand?

It seems like a totally sufficient solution given that this is a page a user only hits once (if ever since it can also be done via `fleetctl`).

noahtalerman:
Got it! Comments are added. That now makes sense that one would attempt to first understand the code in the diff before being presented with the PR description. And for someone who's in the code there will be no PR description.

Re to user only hitting page once if that, sweet!

zwass (approved):
Thanks for adding the comments!

